PR #12, opened 2025-10-23
Read surface topography from Elmer-Xios output

ClemensSchannwell:
In order to read the surface topography from an actual Elmer/Ice simulation, I included the option to `--netcdf-mesh-unstructured`. Since this surface topography uses the library `XIOS`  to write the file, it lives on the same grid as Elmer. This simplifies the reading of the file as the surface topography comes in the same order as the Elmer mesh. 
I think this feature will be helpful in the case of initialisation, because we can read the same surface topography in with which Elmer is initialised and also if we decide that we do not need to update the surface topography within a EBFM run, we can simply read it it at the start of the next simulation. 
I also wanted to write some model output for easy inspection using `ncview` or similar, but while it works in principle, this is still in its early stage. This does not hold up further development of the `EBFM-dummy` but would be good to have to check whether results from EBFM are physically plausible. 
I made some test with the `feature_greenland` branch and all previous documented running commands of the model still work as before (as far as I can tell at least).

To test the new model on levante, you can run: 

```sh
mpirun -np 1 ebfm --elmer-mesh /work/bm1514/data/ebfm_dummy/mesh/MESH --netcdf-mesh-unstructured /work/bm1514/data/elmer-xios-output/restart.nc
```

The `elmer-xios` file should be accessible for all project members on levante.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,9 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Support new input mesh format. EBFM now accepts Elmer/Ice mesh file for xy-coordinates and separate unstructured NetCDF elevation file obtained from XIOS. To use this feature please provide `--elmer-mesh` together with the new option `--netcdf-mesh-unstructured`. See https://github.com/EBFMorg/EBFM/pull/12.
 * Use `setuptools_scm` as backend for `--version` information. See https://github.com/EBFMorg/EBFM/pull/46.
-* Remove `pathlib` from requirements, because this can lead to a bug. https://github.com/EBFMorg/EBFM/pull/48
+* Remove `pathlib` from requirements, because this can lead to a bug. https://github.com/EBFMorg/EBFM/pull/48.
 
 # v0.1.0
 


src/ebfm/INIT.py
```diff
@@ -206,15 +206,18 @@ def init_grid(grid, io, args: Namespace):
             mesh: Mesh = read_elmer_mesh(mesh_root=args.elmer_mesh)
 
         grid["x"], grid["y"] = mesh.x_vertices, mesh.y_vertices
-        grid["z"] = read_dem(args.netcdf_mesh, grid["x"], grid["y"])
+        if args.netcdf_mesh:
+            grid["z"] = read_dem(args.netcdf_mesh, grid["x"], grid["y"])
+            grid["lat"] = np.zeros_like(grid["x"]) + 75  # test values!
+            grid["lon"] = np.zeros_like(grid["x"]) + 320  # test values!
+        if args.netcdf_mesh_unstructured:
+            grid = read_dem_xios(args.netcdf_mesh_unstructured, grid)
```
BenjaminRodenberg: See above. I would rather differentiate based on `grid["input_type"]`. I have the following design in my mind: We only ready `args.netcdf_mesh` and so on once in order to determine the correct `GridInputType` and store it. This also gives us the possibility to check for faulty input early and at a single place. As soon as the `GridInputType` is derived from `args`, we user input processing is done and we can start the actual grid reading with the correct strategy.

src/ebfm/LOOP_write_to_file.py
```diff
@@ -7,7 +7,7 @@
 from netCDF4 import Dataset, date2num
 
 
-def main(OUTFILE, io, OUT, grid, t, time, C):
+def main(OUTFILE, io, OUT, grid, t, time, C, gridtype):
```
BenjaminRodenberg: Is it possible to use `grid["input_type"]` here? Introducing a new type is no problem in my opinion. Extending the interface is something I would only do if necessary.

src/main.py
```diff
@@ -252,19 +259,24 @@ def main():
             logger.debug("Done.")
             logger.debug("Received the following data from Elmer/Ice:", data_from_elmer)
 
-            IN["h"] = data_from_elmer["h"]
-            grid["z"] = IN["h"][0].ravel()
-            # TODO add gradient field later
+            # IN['h'] = data_from_elmer['h']
+            # grid['z'] = IN['h'][0].ravel()
```
BenjaminRodenberg: I assume we need to revert this in order to not break existing functionality?

Why did you have to comment it out in the first place? Is this something that is easy to fix or should we add a guard around this part? I.e.

```python
if grid["grid_type"] == GridInputType.ELMERXIOS:
    logger.warning(f"Grid type {grid["grid_type"]} does not support outputting IN["h"] and grid["z"]. Ignoring these fields.")  # todo: find a fix if necessary
else:
    # usual behavior
```
    
```

File: src/ebfm/INIT.py [hunk 4]
```diff
@@ -338,7 +341,27 @@ def init_grid(grid, io, args: Namespace):
         grid["slope_gamma"][(grid["slope_x"] > 0) & (grid["slope_y"] == 0)] = np.pi / 2
         grid["slope_gamma"][(grid["slope_x"] < 0) & (grid["slope_y"] == 0)] = -np.pi / 2
         grid["slope_gamma"] = -grid["slope_gamma"]
+    elif grid["input_type"] is GridInputType.ELMER_XIOS:
```
BenjaminRodenberg: Does this work?

It's generally a good idea to make sure that at least one of the cases in this chain of `if-elif`-statements is entered. I always append a final `else: raise Exception(f"Cannot handle unexpected {grid["input_type"]=}")`. This will make sure that the program exits.

BenjaminRodenberg (changes requested):
I reviewed the design of this PR and provided some general comments.

I can also take care of the implementation of (some of) these parts if you want. Please just let me know in the comments.

BenjaminRodenberg:
I performed another merge of `main` into this PR. I will push in a moment. Note that there is a good chance that the merge broke something. This should not be dramatic but just in case: The latest known working commit of this PR is [0b14bf3](https://github.com/EBFMorg/EBFM/pull/12/commits/0b14bf3fea4218e1be13f3b91722a7551bd27cff).

src/ebfm/grid.py
```diff
@@ -6,6 +6,14 @@
...
-    CUSTOM = "custom"  # Elmer/Ice mesh file for xy-coordinates and separate NetCDF elevation file
-    ELMER = "elmer"  # Elmer/Ice mesh file with elevation in z-coordinate
+    # .mat grid file with elevation
+    MATLAB = "matlab"
+
+    # Elmer/Ice mesh file for xy-coordinates and separate NetCDF elevation file
+    CUSTOM = "custom"
+
+    # Elmer/Ice mesh file for xy-coordinates and separate unstructured NetCDF elevation file obtained from XIOS
+    XIOS_CUSTOM = "xios_custom"
```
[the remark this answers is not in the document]
ClemensSchannwell: Yes, then please change.

src/ebfm/LOOP_write_to_file.py
```diff
@@ -141,12 +141,17 @@ def save_netCDF_file():
             # Create NetCDF file
             nc_filepath = os.path.join(io["outdir"], "model_output.nc")
             io["nc_file"] = Dataset(nc_filepath, "w", format="NETCDF4")
-
-            # Define dimensions
-            io["nc_file"].createDimension("time", None)  # Unlimited time dimension
-            io["nc_file"].createDimension("y", grid["x_2D"].shape[0])  # 2D grid rows
-            io["nc_file"].createDimension("x", grid["x_2D"].shape[1])  # 2D grid columns
-            io["nc_file"].createDimension("nl", grid["nl"])  # Vertical layers for `sub` variables
+            if gridtype == "unstructured":
```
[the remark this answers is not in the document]
BenjaminRodenberg: Should be resolved.

src/ebfm/config.py
```diff
@@ -100,6 +100,10 @@ def __init__(self, args: Namespace):
             self.grid_type = GridInputType.CUSTOM
             self.mesh_file = args.elmer_mesh
             self.dem_file = args.netcdf_mesh
+        elif args.netcdf_mesh_unstructured and args.elmer_mesh:
+            self.grid_type = GridInputType.ELMERXIOS
+            self.mesh_file = args.elmer_mesh
+            self.dem_file = args.netcdf_mesh_unstructured
```
BenjaminRodenberg:
```suggestion
        elif args.netcdf_mesh_unstructured and args.elmer_mesh:
            self.grid_type = GridInputType.ELMERXIOS
            self.mesh_file = args.elmer_mesh
            self.dem_file = args.netcdf_mesh_unstructured
            self.is_unstructured = True
```

src/main.py
```diff
@@ -126,20 +126,27 @@ def main():
...
         default="1-Jan-1979 00:00",
     )
 
     time_group.add_argument(
         "--end-time",
         type=str,
-        help="End time of the simulation in format 'DD-Mon-YYYY HH:MM', e.g., '02-Jan-1979 09:00'.",
-        default="2-Jan-1979 09:00",
+        help="End time of the simulation in format 'DD-Mon-YYYY HH:MM'",
+        default="2-Jan-1979 00:00",
```
BenjaminRodenberg: I did another merge of `main` into this PR. The default values are quite arbitrary. So I'm fine to just change the default to the value you are using if this saves you some typing.

You can also use the new options to provide different values.

PR #20, opened 2025-11-07
Package the project

BenjaminRodenberg:
* Introduce `pyproject.toml`
* Turn into installable application that is executed as `ebfm`
* Allow checking the version of the application
* Update instructions in `README.md`

Closes #18 
Closes #6

PR #22, opened 2025-11-10
Make YAC an optional dependency

BenjaminRodenberg:
Depends on #20 

Allows to install EBFM without YAC

* Make `import coupling` conditional
* Only require depency `yac` if specified during installation

## Technical

* Refactoring: Introduced `coupler.py` that does not depend on `import yac` to provide interface and only load actual implementation `coupling.py` if `yac` is available.

## Todo

- [x] Update docs
- [x] Try to improve `--help` regarding optional dependencies.
- [x] ~~Add `example` files in package and update instructions.~~ https://github.com/wardvp/EBFM/issues/25


README.md
```diff
@@ -15,14 +15,30 @@ Using YAC this model can be coupled to other models.
...
+
+```sh
+pip install git+https://github.com/wardvp/EBFM.git
+```
+
+Please test your installation by running `ebfm --help` to print the help message and `ebfm --version` to print the installed version.
+
+## Running
+
+After installation a basic, uncoupled simulation can be run with the following command:
```
matthias-ibach: Since I did not clone this repository, I stumbled upon not having the `examples` folder in my directory and, thus, could not run the basic uncoupled example. 
You could possibly add that you need the examples folder from the repo. But I can understand if this is too trivial.

README.md
```diff
@@ -15,14 +15,30 @@ Using YAC this model can be coupled to other models.
...
 
 ```sh
-python3 src/dummy.py --matlab-mesh examples/dem_and_mask.mat
```
matthias-ibach: This simple command worked.
I have not tried the examples with ELMER meshes or netCDF-meshes.

Would it make sense to include some simple Elmer or Elmer Mesh with Elevation data from NetCDF for testing purposes? Or for dummy example purposes? Or is it sufficient to point towards the EBFM-Elmer-Icon-dummy repository?

PR #24, opened 2025-11-11
Update README with alternative pre-commit installation

BenjaminRodenberg:
Added alternative installation method for pre-commit using pip.


ClemensSchannwell (approved):
Yes, this works. Thanks!

PR #23, opened 2025-11-10
Overhaul workflow for preparing venv with YAC

matthias-ibach:

Issue #10, opened 2025-10-17
Add basic CI


BenjaminRodenberg:
I would start with something simple. Basic features:

- [x] Minimal format checking (i.e. PEP8) -> #14 
- [x] Automation of packaging and distribution -> #19 
- [x] Automation of setting version from tags -> #20 
- [x] Minimal testing (probably without YAC and in an uncoupled setup because this would be quite involved) -> #15

PR #30, opened 2025-11-12
Rename `dummy.py` to `main.py`

### LINKED ISSUE #4: Rename `src/dummy.py`
Similar motivation as in #3. `main.py` would be an obvious candidate.

Important: This change will also require some updates of documentation and tooling.

BenjaminRodenberg:
Closes #4 

There are some places in the documentation where there is still `src/dummy.py` mentioned but they will be fixed in #23.


matthias-ibach (commented):
I did a simple test in #10 and it seemed to work.
From my side, this PR is ready to go.

PR #41, opened 2025-11-24
Refactor Coupler class design to use strategy pattern

BenjaminRodenberg:
* Refactoring of `Coupler` class (and derived classes).
* Follows Strategy pattern more closely. 
* Simplifies `main.py` (esp. avoids conditionals)

### DIFF
--- a/src/couplers/yacCoupler.py
+++ b/src/couplers/yacCoupler.py
@@ -5,7 +5,6 @@
 import yac
 import numpy as np
 from dataclasses import dataclass
-from collections import namedtuple
 
 from ebfm import logging
 
@@ -29,12 +28,19 @@ field {name}:
 
 @dataclass
 class FieldDefinition:
-    name: str
+    name: str  # name of the field
     exchange_type: yac.ExchangeType = None  # optional for consistency checks by model configuration
     metadata: str = None  # optional to allow model providing metadata
 
 
+@dataclass
+class Timestep:
+    value: str  # value of the timestep in specified format
+    format: yac.TimeUnit  # format of the timestep value
+
+
 # TODO: Get hard-coded data below from dummies/EBFM/ebfm-config.yaml
+# TODO: Move Timestep inside FieldDefinition (would allow having different timesteps per field)
 all_field_definitions = {
     Component.elmer_ice: [
         FieldDefinition(
@@ -224,14 +230,13 @@ class YACCoupler(Coupler):
         self.grid.set_global_index(grid.vertex_ids, yac.Location.CORNER)
         self.corner_points = self.grid.def_points(yac.Location.CORNER, grid.lon, grid.lat)
 
-    def _add_couples(self, time: Dict[str, float]):
+    def _add_couples(self, timestep: Timestep):
         """
         Adds coupling definitions to the Coupler interface.
 
-        @param[in] time dictionary with time parameters, e.g. {'tn': 12, 'dt': 0.125}
+        @param[in] timestep Timestep object representing the timestep size of the fields
         """
-
-        self._construct_coupling_pre_sync(time)
+        self._construct_coupling_pre_sync(timestep)
 
         self.interface.sync_def()
 
@@ -239,11 +244,11 @@ class YACCoupler(Coupler):
 
         self.interface.enddef()
 
-    def _construct_coupling_pre_sync(self, time: Dict[str, float]):
+    def _construct_coupling_pre_sync(self, timestep: Timestep):
         """
         Constructs the coupling interface with YAC.
 
-        @param[in] time dictionary with time parameters, e.g. {'tn': 12, 'dt': 0.125}
+        @param[in] timestep Timestep object representing the timestep size of the fields
         """
 
         assert self.fields == {}, "Coupling fields have already been constructed."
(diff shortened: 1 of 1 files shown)


src/ebfm/LOOP_EBM.py
```diff
@@ -14,7 +14,7 @@
 )
 from ebfm import LOOP_EBM_SWout, LOOP_EBM_insolation
 
-from coupler import Coupler
+from couplers import Coupler
```
BenjaminRodenberg: I could also avoid these breaking changes by introducing additional wrapper modules. But this feels like a bit too much.

src/couplers/base.py
```diff
@@ -0,0 +1,159 @@
...
+            "### Design recommendations ###\n"
+            "It is recommended to follow the strategy pattern and initialize your Coupler as a DummyCoupler, "
+            "YACCoupler or similar to avoid general has_coupling checks since this usually leads to a better design.\n"
+            "Alternatively, you can check for the type of your coupler. A DummyCoupler never has a coupling and you "
+            "can use this to perform checks similar to has_coupling.\n"
+            "##############################",
+        )
+        return_value = self._couple_to_elmer_ice or self._couple_to_icon_atmo
+        logger.debug(f"Component {self.component_name} has_coupling={return_value}")
+        return return_value
```
BenjaminRodenberg: I will replace `has_coupling` as suggested in this deprecation message in a follow-up PR. Similar for other cases where `deprecation` is used.

src/couplers/base.py
```diff
@@ -0,0 +1,159 @@
...
+            )
+            grid = grid["mesh"]
+
+        self._setup(grid, time)
+
+    def _setup(self, grid: Grid, time: Dict[str, float]):
+        """
+        Implementation of Coupler.setup
+
+        @param[in] grid Grid used by EBFM where coupling happens
```
k202077: see above

src/couplers/base.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        deprecation(
+            logger, 'Coupler.exchange_elmer_ice(data) is deprecated, use Coupler.exchange("elmer_ice", data) instead.'
+        )
+        return self.exchange("elmer_ice", data_to_exchange)
+
+    def exchange_icon_atmo(self, data_to_exchange: Dict[str, np.array]) -> Dict[str, np.array]:
+        """Exchange data with ICON atmosphere component
+
+        @param[in] put_data dictionary of field names and their data to be exchanged with ICON atmosphere component
```
k202077: see above

src/couplers/yacCoupler.py
```diff
@@ -101,10 +102,10 @@ def construct_coupling(self, time: Dict[str, float]):
         @param[in] time dictionary with time parameters, e.g. {'tn': 12, 'dt': 0.125}
```
k202077: private method?

k202077 (commented):
duplicated code:
```
        for field_definition in field_definitions:
            field = yac.Field.create(
                field_definition.name,
                self.component,
                self.corner_points,
                collection_size,
                timestep.value,
                timestep.format,
            )
            self.interface.def_field_metadata(
                field.component_name, field.grid_name, field_definition.name, field_definition.metadata.encode("utf-8")
            )
            if field_definition.exchange_type == yac.ExchangeType.SOURCE:
                self.source_fields[field_definition.name] = field
            elif field_definition.exchange_type == yac.ExchangeType.TARGET:
                self.target_fields[field_definition.name] = field
```
Could be turned into a private method that takes a single field definition (`exchange_type`, `name`, and `metadata`) as an argument. This would make method `construct_coupling_elmer_ice` and `construct_coupling_icon_atmo` much shorter, because you could add the fields directly without having to store them in `field_defintions`.

First check in `construct_coupling_post_sync` seems redundant and second check out also be applied to source fields.

In `field_information`, a field does not have to have metadata. You can query whether metadata was defined for a field.

src/couplers/base.py
```diff
@@ -0,0 +1,159 @@
...
+        @param[in] component_name name of the component to check coupling for
+
+        @returns True if coupling to the specified component is enabled, False otherwise
+        """
+        if component_name == "icon_atmo":
+            return self._couple_to_icon_atmo
+        elif component_name == "elmer_ice":
+            return self._couple_to_elmer_ice
+        else:
+            raise ValueError(f"Unknown component name '{component_name}' for coupling check.")
```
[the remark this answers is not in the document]
BenjaminRodenberg: I'm already developing a more generic method. 100% agree that enums are much better. Will push it together with other changes from this review.

src/couplers/base.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        Setup the coupling interface
+
+        Performs initialization operations after init and before entering the
+        time loop
+
+        @param[in] grid Grid used by EBFM where coupling happens
+        @param[in] time dictionary with time parameters, e.g. {'tn': 12, 'dt': 0.125}
+        """
+        print(type(grid))
```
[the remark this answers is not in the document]
BenjaminRodenberg: Added this by accident. Will be removed.

src/couplers/base.py
```diff
@@ -0,0 +1,159 @@
...
+        raise NotImplementedError("Coupler is an abstract base class and cannot be instantiated directly.")
+
+    def setup(self, grid: Grid | Dict, time: Dict[str, float]):
+        """
+        Setup the coupling interface
+
+        Performs initialization operations after init and before entering the
+        time loop
+
+        @param[in] grid Grid used by EBFM where coupling happens
```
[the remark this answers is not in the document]
BenjaminRodenberg:
```suggestion
        @param[in] grid Grid used by EBFM where coupling fields are defined
```

BenjaminRodenberg (commented):
I already have some changes ready that take care of some parts of this review. Will push them probably in the next few hours.

src/couplers/yacCoupler.py
BenjaminRodenberg: Done. 6eaae6a699487066b457b5c35680b612026700df

Issue #11, opened 2025-10-17
Add shading functionality


wardvp:
Add the following functions/code to enable local calculations of shading:
1) a function before the time-loop that can determine maximum grid angles (for uncoupled runs)
2) add YAC instructions to retrieve maximum grid angle arrays & slope information from Elmer/Ice
3) rewrite the shading function in the time-loop to determine shading based on the solar elevation angle and the maximum grid angle.

BenjaminRodenberg:
Since this feature is not crucial for the first release I would suggest to add it at a later point in time. The release process is quite simple and we can do a new release as soon as this feature is ready.

PR #44, opened 2025-11-25
Add CouplingConfig and GridConfig

BenjaminRodenberg:
This PR overhauls the configuration of the coupler and the grid. The overall design now uses the `args` from the CLI parser and creates `*Config` objects after performing some consistency checks. This avoids using `args` at later stages of executions and developers can assume that the configuration provided by `*Config` is valid.

### DIFF
--- a/src/ebfm/INIT.py
+++ b/src/ebfm/INIT.py
@@ -184,7 +184,7 @@ def init_grid(grid, io, config: GridConfig):
         grid["mask"] = np.ones_like(grid["x"])  # treats every grid cell as glacier
         grid["gpsum"] = np.sum(grid["mask"] == 1)  # number of modelled grid cells
         grid["mesh"] = mesh
-        grid["has_shading"] = False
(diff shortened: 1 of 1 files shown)


src/ebfm/LOOP_EBM_insolation.py
```diff
@@ -71,7 +69,7 @@ def main(grid, time2, OUT):
         np.sin(lat_rad) * np.sin(d_rad) + np.cos(lat_rad) * np.cos(d_rad) * np.cos(h_rad)
     )  # SOURCE: Iqbal (1983)
 
-    if grid["input_type"] is GridInputType.MATLAB:  # only use shading routine in uncoupled runs on a regular grid
+    if grid["has_shading"]:
         yl = grid["Ly"]
         xl = grid["Lx"]
```
BenjaminRodenberg: Instead of checking for the `GridInputType` I would generally check whether shading is supported.

BenjaminRodenberg (commented):
@wardvp maybe some parts of this PR are interesting for #11 ? I'm generally trying to move towards a slightly more object oriented approach here.

PR #46, opened 2025-11-25
Use `setuptools_scm` for more informative development version

BenjaminRodenberg:
Using `setuptools_scm` for the following reasons:

* More flexible w.r.t development releases and tagging. I observed problems when using tags like `v0.1.0-dev1`. Here, `setuptools_scm` seems to be more robust
* More informative output if no git history is available. This is especially useful if fetching only the latest commit (i.e. in the CI). `setuptools-git-versioning` defaults to `0.0.1` in this case which is misleading. The default of `setuptools_scm` is more useful for logging, debugging, user support, etc.

# Author's Todos

- [x] I added an entry under *develop* in the [CHANGELOG.md](https://github.com/EBFMorg/EBFM/blob/main/CHANGELOG.md) (may be skipped for minor changes not obervable for the user)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ SPDX-License-Identifier: CC-BY-4.0
 # develop
 
 * Use `setuptools_scm` as backend for `--version` information. See https://github.com/EBFMorg/EBFM/pull/46.
+* Remove `pathlib` from requirements, because this can lead to a bug. https://github.com/EBFMorg/EBFM/pull/48
 
 # v0.1.0
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,6 @@ dependencies = [
     "numpy >= 2.0",
     "pandas >= 2.0",
     "pandas-stubs >= 2.0",
-    "pathlib",
     "pyproj",
     "scipy",
 ]


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Use `setuptools_scm` as backend for `--version` information. See https://github.com/EBFMorg/EBFM/pull/46.
```
matthias-ibach: I checked how the `setuptools_scm` works locally (by cloning shallow, tagging and checking the version output), and (of course) it seemed to work just fine.
This was just a small exercise I did.

pyproject.toml
```diff
@@ -32,11 +32,12 @@ cpl = [
...
 [build-system]
-requires = ["setuptools", "setuptools-git-versioning"]
+requires = ["setuptools", "setuptools_scm"]
 build-backend = "setuptools.build_meta"
 
-[tool.setuptools-git-versioning]
-enabled = true
+[tool.setuptools_scm]
+version_scheme = "post-release"
+local_scheme = "node-and-date"
```
matthias-ibach: Unfortunately, I do not know too much about other versioning schemes that could be used here, but my fast and limited research said that "post-release" etc. is fine as it is _"...good if the project does many intermediate builds between formal releases."_

PR #58, opened 2025-12-12
Improve time configuration and move to zero-based indexing

BenjaminRodenberg:
This pull-request mainly started as a usability improvement. I think it does not feel very natural if the following call performs in total 12 time steps (each 3h):

```sh
ebfm --start-time "1-Jan-1979 00:00" --end-time "2-Jan-1979 9:00" --matlab-mesh ~/Programming/EBFM_with_YAC/examples/dem_and_mask.mat
EBFM r0 2025-12-12 17:52:50.456 INFO main - main: Starting EBFM version 0.1.0.post16+g08d87a129...
EBFM r0 2025-12-12 17:52:50.456 INFO main - main: Done parsing command line arguments.
EBFM r0 2025-12-12 17:52:50.457 INFO config - __init__: Using non-partitioned grid...
EBFM r0 2025-12-12 17:52:50.461 INFO INIT - read_MATLAB_grid: EBFM: Reading grid data from MATLAB file...
EBFM r0 2025-12-12 17:52:50.641 INFO INIT - init_initial_conditions: EBFM: Initialize from manually set conditions...
EBFM r0 2025-12-12 17:52:50.656 INFO main - main: Entering time loop...
Year: 1979   Month:  1   Day:  1   Hour:  0
EBFM r0 2025-12-12 17:52:50.656 INFO main - main: Time step 1 of 12 (dt = 0.125 days)
Year: 1979   Month:  1   Day:  1   Hour:  3
EBFM r0 2025-12-12 17:52:54.575 INFO main - main: Time step 2 of 12 (dt = 0.125 days)
...
Year: 1979   Month:  1   Day:  2   Hour:  6
EBFM r0 2025-12-12 17:53:31.235 INFO main - main: Time step 11 of 12 (dt = 0.125 days)
Year: 1979   Month:  1   Day:  2   Hour:  9
EBFM r0 2025-12-12 17:53:35.290 INFO main - main: Time step 12 of 12 (dt = 0.125 days)
Boot file saved to /home/benjamin/Programming/EBFM_with_YAC/Reboot/boot_final.nc
EBFM r0 2025-12-12 17:53:41.317 INFO main - main: Time loop completed.
EBFM r0 2025-12-12 17:53:41.318 INFO main - main: Closing down EBFM.
```

I would rather see `--end-time "2-Jan-1979 12:00"` to get this behavior. I.e., `--end-time` refers to the end of the simulation time and not to the beginning of the final time step.

This is linked tightly to how one counts time steps. I would use the following rules:

* (t_end - t_start) / dt gives the number of time steps tn (here: 12)
* t_start has the index 0, t_end (if needed) the index tn (here: 12)
* Print statements should read naturally. I would not write something like "Time step 0 of 12" but "Time step 1 of 12". Time step 1 brings us from time 0 (t_start) to time 1 (t_start + dt)

## More Context

See https://gitlab.dkrz.de/TerraDT/ebfm_dummy/-/merge_requests/4

# Author's Todos

- [x] I added an entry under *develop* in the [CHANGELOG.md](https://github.com/EBFMorg/EBFM/blob/main/CHANGELOG.md) (may be skipped for minor changes not obervable for the user)


src/ebfm/LOOP_write_to_file.py
```diff
@@ -55,13 +57,13 @@ def main(OUTFILE, io, OUT, grid, t, time, C):
...
         # Initialize TEMP storage
-        if (t - 1) % io["freqout"] == 0:
+        if t % io["freqout"] == 0:
             OUTFILE.setdefault("TEMP", {})
             OUTFILE["TEMP"][varname] = np.zeros_like(temp_long)
 
         # Handle type: sample, mean, or sum
         if var_type == "sample":
-            if (t + io["freqout"] // 2) % io["freqout"] == 0:
+            if (t + 1 + io["freqout"] // 2) % io["freqout"] == 0:  # TODO: correct?
```
BenjaminRodenberg: My understanding: If `io["freqout"] == 1` means that we want to write every single time step, `io["freqout"] == 2` every second time step and so on. Adding `+ io["freqout"] // 2` shifts the point where the sampling is triggered from the beginning to the middle of one period.

Example: If I set `io["freqout"] = 3` this would result in the bold time steps being written to `OUTFILE`: 0 **1** 2 | 3 **4** 5 | ... ; for even numbers are somewhat it is not really clear where the middle value is supposed to be.

Generally, I think we should clarify the expected behavior and consider moving the condition here into a function like `is_middle_t_of_every_n(...) -> bool` (open for suggestions regarding the name).

src/ebfm/LOOP_climate_forcing.py
```diff
@@ -68,7 +70,7 @@ def main(C, grid, IN, t, time, OUT, cpl: Coupler) -> tuple[dict, dict]:
     # Time since last snowfall event
     snowfall_mask = (IN["snow"] / (time["dt"] * C["dayseconds"])) > C["Pthres"]
     OUT["timelastsnow"][snowfall_mask] = time["TCUR"]
-    if t == 1:
+    if is_first_time_step(t, time):
```
k202077: To get this kind of behaviour, I would use a static boolean which is initialised to true and set to false during the first iteration.

BenjaminRodenberg: I did not want to change too many things regarding the logic in this PR.

But this part looks problematic to me since `LOOP_climate_forcing.main` overwrites what has been set in `INIT.main` (even if there is a restart file).

I would prefer to keep things as they are here. A potential bugfix would entirely remove the need for `is_first_time_step` here.

PR #60, opened 2025-12-16
Implement faster shading calculation based on look-up tables

### LINKED ISSUE #11: Add shading functionality

wardvp:
Replace the original shading function that walks towards the sun every time-step with a method that uses a look up table of maximum grid angles. A grid cell is shaded when the solar elevation angle is below the maximum grid angle. The new shading method implies that all grid cells can be modeled independently, i.e. the code can be run in parallel. 

Tasks:
- Generate maximum grid angle look-up table for a set of azimuth angles (used in GridInputType.MATLAB case)
- Add functionality in LOOP_EBM_insolation that reads from the look-up table using the nearest azimuth angle for the time of the day.


.gitignore
wardvp: The only way for me to push changes related to the new shading code was to put these into .gitignore. Not sure what these are though.

src/ebfm/INIT.py
```diff
@@ -328,6 +328,79 @@ def init_grid(grid, io, config: GridConfig):
...
+
+        # loop over the azimuth angles to determine gridded maximum grid angles per angle
+        grid["maxgridangle"] = np.zeros((grid["gpsum"], grid["nr_az_steps"]), dtype=np.float64)
+        for n in range(grid["nr_az_steps"]):
+            az = np.full(int(grid["gpsum"]), grid["az_array"][n], dtype=float)
+
+            # calculate step sizes (ddx, ddy) in x- and y-directions for all azimuth angles
+            ddx = np.empty_like(az, dtype=float)
+            ddy = np.empty_like(az, dtype=float)
+            m = az <= -0.75 * np.pi
```
BenjaminRodenberg: What is `m`?

src/ebfm/INIT.py
```diff
@@ -328,6 +328,79 @@ def init_grid(grid, io, config: GridConfig):
...
+
+                iv = i[inbound]
+                jv = j[inbound]
+
+                dx = grid["x_2D"][iv, jv] - grid["x"][inbound]
+                dy = grid["y_2D"][iv, jv] - grid["y"][inbound]
+                distance = np.hypot(dx, dy)  # walk distance from start to target
+
+                dz = grid["z_2D"][iv, jv] - grid["z"][inbound]
+                grid_angle = np.arctan(dz / distance)  # grid angle from start to target
```
BenjaminRodenberg: Put this into a function?

`grid_angle = compute_angle(grid, iv, ij, inbound)` ?

You could additionally offer a function `grid_angle_proxy = compute_angle_for_max(grid, iv, ij, inbound)`.

BenjaminRodenberg:
```suggestion
                grid_angle_proxy = np.arctan(dz**2 / distance_sq)  # grid angle from start to target
```
If we want to use squared distance we also have to use the squared `dz`.

src/ebfm/INIT.py
```diff
@@ -328,6 +328,79 @@ def init_grid(grid, io, config: GridConfig):
...
+
+                max_angle[inbound] = np.maximum(max_angle[inbound], grid_angle)  # update max grid angle when needed
+
+                active &= (j >= 0) & (j < yl) & (i >= 0) & (i < xl)  # continue walk until domain edge is reached
+
+                count += 1
+
+            # fill lookup table with maximum grid angles for all cells (dimension 1) and azimuth angle (dimension 2)
+            grid["maxgridangle"][:, n] = max_angle
```
BenjaminRodenberg: If we use the squared distance above we would have to apply a correction here:

```suggestion
            grid["maxgridangle"][:, n] = np.arctan(np.sqrt(np.tan(max_angle_proxy)))
```

This costs more than a `sqrt`, but we are saving a lot of `sqrt`s above. So it should pay off. But: let's not do it if it is not measurable.

src/ebfm/LOOP_EBM_insolation.py
```diff
@@ -141,10 +153,26 @@ def main(grid, time2, OUT):
...
+        # Shading happens when the solar elevation angle <= maximum grid angle
+        temp = grid["maxgridangle"]
+        OUT["shade"] = elevationangle <= temp.flat[ind_az]
+
+        # For now: assume flat surface and no shading by surrounding terrain (no longer needed!)
+        # OUT["shade"] = np.zeros_like(grid["x"])
+        # OUT["shade"][elevationangle < 0] = 1
+
+    OUT["shade_2D"] = zeros_like(grid["x_2D"])
+    OUT["shade_2D"].flat[grid["ind"]] = OUT["shade"]
```
BenjaminRodenberg: Unused?

Instead of storing and duplicating these flattened `_2D` data you could also implement a small helper function similar to the following:

```
shade_2D = flatten(grid, "shade")
```

This is a general thing because I think I saw this pattern at quite a few places.

BenjaminRodenberg (commented):
I cannot say a lot about the algorithm but generally the implementation looks good. 

Main comments are on readability and coding style. If you want, I can also take over here and implement the suggested changes (I will only need a few clarifications in my review comments).

Regarding testing: The matlab example covers the shading. We are still lacking a proper test where we compare the actual results to reference data. But this is something I can also take care of. I assume that the current output when I run the Matlab example is correct and could be used as a reference result to be able to detect future regressions. (We especially need this before starting anything in the direction of performance optimization)

src/ebfm/INIT.py
```diff
@@ -244,7 +244,7 @@ def init_grid(grid, io, config: GridConfig):
         mask_2D = input_data["mask"][0][0]
 
         # Determine domain extent
-        grid["has_shading"] = True
+        grid["classical_shading"] = False
```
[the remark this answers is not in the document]
wardvp: The classical shading refers to the original shading method, which could be a useful reference for testing (it is in fact more slightly more accurate than the new shading method which only approximates the azimuth angle). But indeed it is not used by any of the grid types right now, so maybe it does not have to remain as a condition.

BenjaminRodenberg (commented):
I did a merge of `main` into this branch and performed a few of the simpler changes I suggested earlier.

I think we still should move some of the shading functionality into functions (maybe even an independent module) to make the code easier to read.

src/ebfm/core/INIT.py
```diff
@@ -328,6 +328,79 @@ def init_grid(grid, io, config: GridConfig):
...
+        # loop over the azimuth angles to determine gridded maximum grid angles per angle
+        grid["maxgridangle"] = np.zeros((grid["gpsum"], grid["nr_az_steps"]), dtype=np.float64)
+        for n in range(grid["nr_az_steps"]):
+            az = np.full(int(grid["gpsum"]), grid["az_array"][n], dtype=float)
+
+            # calculate step sizes (ddx, ddy) in x- and y-directions for all azimuth angles
+            ddx = np.empty_like(az, dtype=float)
+            ddy = np.empty_like(az, dtype=float)
+            m = az <= -0.75 * np.pi
+            ddx[m] = -np.tan(np.pi + az[m])
```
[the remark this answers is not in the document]
wardvp: Yes, it seems correct to me: az<-0.75pi means it is walking towards the SSW, i.e. negative steps in the W-E direction (ddx) and positive steps in the N-S (ddy). Please note that eastward means ddx positive, westward means ddx negative, northward means ddy negative and southward means ddy positive. ddx and ddy are determining in which direction we are walking and at what 'speed'.

src/ebfm/core/INIT.py
```diff
@@ -328,6 +328,79 @@ def init_grid(grid, io, config: GridConfig):
...
+        grid["az_array"] = -np.pi * (-1.0 + 2.0 * (np.arange(1, grid["nr_az_steps"] + 1) / grid["nr_az_steps"]))
+
+        xl, yl = grid["x_2D"].shape
+
+        # loop over the azimuth angles to determine gridded maximum grid angles per angle
+        grid["maxgridangle"] = np.zeros((grid["gpsum"], grid["nr_az_steps"]), dtype=np.float64)
+        for n in range(grid["nr_az_steps"]):
+            az = np.full(int(grid["gpsum"]), grid["az_array"][n], dtype=float)
+
+            # calculate step sizes (ddx, ddy) in x- and y-directions for all azimuth angles
```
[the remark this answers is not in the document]
wardvp: I have now created this function.

src/ebfm/core/INIT.py
```diff
@@ -328,6 +328,79 @@ def init_grid(grid, io, config: GridConfig):
...
+            ddx = np.empty_like(az, dtype=float)
+            ddy = np.empty_like(az, dtype=float)
+            m = az <= -0.75 * np.pi
```
wardvp: I tried to make it clearer now what ddx and ddy are with inline comments and some variable renaming. It is just setting horizontal steps in the N-S (ddy) and E-W (ddx) direction. These are used to jump to the right grid cells during the walk towards the azimuth later on.

src/ebfm/core/LOOP_write_to_file.py
```diff
@@ -54,7 +54,9 @@ def main(OUTFILE, io, OUT, grid, t, time):
     # Update OUTFILE.TEMP with variables to be stored
     for entry in OUTFILE["varsout"]:
         varname, var_type = entry[0], entry[2]
-        temp_long = np.float64(OUT[varname])
+        # "shade" is written as a fraction but sourced from the boolean OUT["is_shaded"]
+        source_key = "is_shaded" if varname == "shade" else varname
+        temp_long = np.float64(OUT[source_key])
```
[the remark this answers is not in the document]
wardvp: Ok, yes, that makes sense.

src/ebfm/core/INIT.py
```diff
@@ -328,6 +328,79 @@ def init_grid(grid, io, config: GridConfig):
...
+
+                iv = i[inbound]
+                jv = j[inbound]
+
+                dx = grid["x_2D"][iv, jv] - grid["x"][inbound]
+                dy = grid["y_2D"][iv, jv] - grid["y"][inbound]
+                distance = np.hypot(dx, dy)  # walk distance from start to target
+
+                dz = grid["z_2D"][iv, jv] - grid["z"][inbound]
+                grid_angle = np.arctan(dz / distance)  # grid angle from start to target
```
wardvp: Ok, done.

src/ebfm/core/INIT.py
```diff
@@ -328,6 +328,79 @@ def init_grid(grid, io, config: GridConfig):
...
         grid["slope_gamma"][(grid["slope_x"] < 0) & (grid["slope_y"] == 0)] = -np.pi / 2
         grid["slope_gamma"] = -grid["slope_gamma"]
+
+        # -----------------------------------------------------------------------------------------------------
+        # Pre-compute maximum grid elevation angle for various azimuth angles (needed for shading calculation)
+        # -----------------------------------------------------------------------------------------------------
+        grid["nr_az_steps"] = 24  # number of azimuth angles (e.g. 24 = 1 per hour)
+
+        # azimuth angles in radians from -pi to +pi with nr_az_steps number of steps
+        grid["az_array"] = -np.pi * (-1.0 + 2.0 * (np.arange(1, grid["nr_az_steps"] + 1) / grid["nr_az_steps"]))
```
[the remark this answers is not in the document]
wardvp: Ok, done.

BenjaminRodenberg (approved):

PR #63, opened 2025-12-19
Introduce FieldSet class

BenjaminRodenberg:
* Offer filtering functionality in FieldSet.
* Provide timestep with field definitions.
* Reorganize and put field constructions inside Field.
* Provide field definitions from outside.

# Author's Todos

### DIFF
--- a/src/couplers/yacCoupler.py
+++ b/src/couplers/yacCoupler.py
@@ -88,14 +88,14 @@ class Field:
         return replace(self, yac_field=yac_field)
 
 
-class FieldCollection:
+class FieldSet:
     """
-    Collection of fields.
+    Set of fields.
 
     Can be used to collect fields and perform filtering operations for components, exchange types, etc.
 
     Example:
-        fields = FieldCollection()
+        fields = FieldSet()
         fields.add(Field(..., exchange_type=yac.ExchangeType.SOURCE))
         fields.add(Field(..., exchange_type=yac.ExchangeType.TARGET))
         source_fields = fields.filter(lambda f: f.exchange_type == yac.ExchangeType.SOURCE)
@@ -103,7 +103,7 @@ class FieldCollection:
 
     def __init__(self, fields: Set[Field] = None):
         """
-        Initialize FieldCollection.
+        Initialize FieldSet.
         """
         self._fields = fields if fields is not None else set()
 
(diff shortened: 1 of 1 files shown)


src/couplers/yacCoupler.py
```diff
@@ -26,97 +26,229 @@
...
+        if self.exchange_type:
+            field_role = yac_interface.get_field_role(yac_field.component_name, yac_field.grid_name, yac_field.name)
+            assert field_role == self.exchange_type, (
+                f"Field '{self.name}' role mismatch: expected '{self.exchange_type}', " f"got '{field_role}'."
+            )
+
+        return replace(self, yac_field=yac_field)
+
+
+class FieldCollection:
```
BenjaminRodenberg: The name collection is used in a different context in YAC. Try to find a different name. E.g, group or simply set.

BenjaminRodenberg: Fixed: [8513b2a](https://github.com/EBFMorg/EBFM/pull/63/commits/8513b2a2e0b1992f25a758605c79279733df56e0)

src/couplers/yacCoupler.py
```diff
@@ -26,97 +26,229 @@
...
+
+@dataclass(frozen=True)
+class Field:
+    """
+    Object for definition of a field to be exchanged via YAC.
+    """
+
     name: str  # name of the field
-    exchange_type: yac.ExchangeType = None  # optional for consistency checks by model configuration
+    component: Component  # component this field belongs to
```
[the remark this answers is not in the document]
BenjaminRodenberg: A renaming makes the situation at least a bit less confusing. See [beb5db1](https://github.com/EBFMorg/EBFM/pull/63/commits/beb5db1b9633770b84bd21836afd2a43f6fcd15d).

I think with the current workflow it is hard to avoid storing the name of the other component in the field and also use it as part of the definition. However, this does not cover cases where one field is being send to multiple components. Defining such a field would be perfectly valid but I believe YAC would complain with the current implementation because we would actually try to define the same field twice in `construct_yac_field`: `Field.coupled_component` does not show up (for good reasons) in the following call:

```py
yac_field = yac.Field.create(
    self.name,
    yac_component,
    corner_points,
    collection_size,
    self.timestep.value,
    self.timestep.format,
)
```

Maybe it is a bad idea to offer `construct_yac_field` as part of `Field` because this enforces a 1-to-1 mapping which actually makes the interface less flexible than it should be. I have the impression having a user interface like `data_from_elmer = coupler.exchange("elmer_ice", data_to_elmer)` could be part of the problem.

@k202077 I think we need to discuss this one more time. I'm wondering how this works in ICON.

PR #69, opened 2026-01-13
Update usage of reader.py

matthias-ibach:
This PR addresses the usage of `reader.py` and adds a small section in the `README.md` to explain how to use it.

# Author's Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Fix and extend `reader.py`, documentation on how to use it and how to obtain required example data. https://github.com/EBFMorg/EBFM/pull/69.
 * Revise folder layout to avoid clutter in `site-packages`. Installing EBFM should now only affect `site-packages/ebfm`. https://github.com/EBFMorg/EBFM/pull/73.
 * Require Python minimum version 3.9. (Planned to increase to 3.10 soon)
 * Clarification how `--start-time` and `--end-time` is interpreted by EBFM. Require that difference of start and end time is a multiple of `--time-step`. https://github.com/EBFMorg/EBFM/pull/58.


README.md
```diff
@@ -151,6 +151,36 @@ Usage example for partitioned mesh:
 ebfm --elmer-mesh examples/MESH/partitioning.128/ --netcdf-mesh examples/BedMachineGreenland-v5.nc --is-partitioned-elmer-mesh --use-part 42
 ```
 
+### Using `reader.py` to prepare an Elmer mesh with elevation data
+
+The helper script `reader.py` can be used to combine an existing Elmer mesh with a DEM NetCDF file by replacing the z‑coordinates in `mesh.nodes` ahead of time. This can be useful if you want to preprocess a mesh once and reuse it for multiple EBFM runs.
+
+For example, you can run the following command from the repository root (`src`):
```
BenjaminRodenberg: This is contradictory. I would rather do it consistently and always write `python3 src/reader.py examples/MESH_with_DEM` (so assuming the user is at the root of the repo and not inside `src`).

Remark: If you are interested in how this works, you could even turn `src/reader.py` into a proper executable like `ebfm` (which maps to `src/main.py`). The entry point for `ebfm` is defined [here](https://github.com/EBFMorg/EBFM/blob/ba2c630593c289584a23c50043d460ff289e4a80/pyproject.toml#L32). However, this is in my opinion not really needed (still nice-to-have) for such a niche tool and I would only do it for the learning.

README.md
```diff
@@ -151,6 +151,36 @@ Usage example for partitioned mesh:
...
+
+```sh
+python3 reader.py ../examples/MESH ../examples/BedMachineGreenland-v5.nc
```
BenjaminRodenberg:
```suggestion
python3 src/reader.py examples/MESH examples/BedMachineGreenland-v5.nc
```

Consistently others below.

README.md
BenjaminRodenberg: What's generally missing and might be slightly irritating: How do you get the files `BedMachineGreenland-v5.nc` and `MESH`? I would not duplicate what we already wrote in https://gitlab.dkrz.de/TerraDT/testcase because this has to be maintained. But maybe add a pointer to the repository?

src/reader.py
```diff
@@ -243,4 +256,8 @@ def fortran_style_sci(x, precision=15):
     y = mesh.y_vertices
     h = read_dem(args.dem, x, y)
 
-    write_dem_as_elmer(mesh, h, args.elmer_mesh / "MESH" / "mesh.nodes", allow_overwrite=True)
+    if outpath != args.elmer_mesh:
+        assert args.elmer_mesh.is_dir(), f"{args.elmer_mesh} is no directory or does not exist."
+        shutil.copytree(args.elmer_mesh, outpath)
+
+    write_dem_as_elmer(mesh, h, outpath / "mesh.nodes", allow_overwrite=True)
```
BenjaminRodenberg:
```suggestion
    write_dem_as_elmer(mesh, h, outpath / "mesh.nodes", allow_overwrite=args.in_place)
```

I would still prefer this, because I find it unintuitive to set `allow_overwrite=True` if we are technically not overwriting the `args.elmer_mesh`. But with the current implementation the way I'm suggesting above does not work. If you want: Feel free to come up with a different approach. But I can also live with the current one.

BenjaminRodenberg (changes requested):

BenjaminRodenberg (approved):
Feel free to merge as soon as the last part with the `allow_overwrite` is integrated.

Issue #55, opened 2025-12-10
Allow running EBFM in parallel


BenjaminRodenberg:
At some point EBFM might turn into a bottleneck. As far as I know, EBFM is [embarrassingly parallel](https://en.wikipedia.org/wiki/Embarrassingly_parallel) (at least if we do not use shading). It should be possible to speed up the computation by using [Numba](https://numba.pydata.org/) or something similar.

I would suggest to not support parallel runs if shading is active and throw an error if a user tries to activate shading during a parallel.

Note: We should take care of #38 first. Additionally, we have to keep in mind that for coupled runs we must do the communicator splitting via YAC(mpi_handshake).

BenjaminRodenberg:
#88 has some potential side-effects. Therefore waiting with this feature for the next release.

matthias-ibach:
PR #105 improves EBFM performance by optimizing the main kernels: `compaction`, `heat_conduction`, `percolation_refreezing_and_storage`, and `layer_merging_and_splitting`. Identified via profiling as the slowest components.

## Results

| MATLAB |     | Total Time | Kernel  <br>Compaction | Kernel   <br>Heat Conduction | Kernel  <br>Percolation | Kernel  <br>Layer-Merging |
| --- | --- | --- | --- | --- | --- | --- |
| Baseline | Numpy | 11.29 s | 1.35739 s | 5.02416 s | 1.66053 s | 0.326749 s |
| Optimization Compaction | Numpy | 10.95 s | 1.28806 s | n.a. | n.a. | n.a. |
|     | Numba | 12.94 s | 0.451926 s | n.a. | n.a. | n.a. |
| Optimization  <br>Heat Conduction | Numpy | 8.19 s | n.a. | 2.63368 s | n.a. | n.a. |
|     | Numby | 6.12 s | n.a. | 0.576871 s | n.a. | n.a. |
| Optimization  <br>Percolation   <br><br/> | Numpy | 10.6 s | n.a. | n.a. | 1.66269 s | n.a. |
|     | Numba | 12.16 s | n.a. | n.a. | 0.403689 s | n.a. |
| Optimization  <br>Layer-Merging | Numpy | 10.5 s | n.a. | n.a. | n.a. | 0.268297 s |
| Optimization  <br>ALL | Numpy | 7.45 s | 1.73546 s | 2.56465 s | 1.49835 s | 0.253038 s |
[...]

PR #105, opened 2026-03-27
i55-performance-optimization

matthias-ibach:
This PR introduces performance optimizations (with refined NumPy improvements and Numba kernels), reproducibility features and a diagnostic tool to EBFM to compare snapshots between runs.

Key Changes:
* Optimized NumPy path: Improved vectorization and reduced memory allocations in core physics loops (`LOOP_SNOW.py`), e.g., better use of `np.where, np.minimum` and pre-computed arrays
* Numba acceleration: Parallel kernels for `compaction`, `heat_conduction` and `percolation` (opt-in via `--with-numba`)
* `--random-seed` fixes random state for consistent results
* `--diagnostics` logs timestep stats; `--dump-reference` saves full model state for comparison
* `tools/compare_snapshots.py` checks model outputs for regression testing

Note: Numba support requires `pip install 'ebfm[performance]'`. All features are optional and backward-compatible.

Some performance results are summarized in issue #55 .


src/ebfm/core/INIT.py
```diff
@@ -487,9 +487,9 @@ def init_initial_conditions(C, grid: GridDict, io, time, init_with_restart_file:
 
         if grid.get("doubledepth", False):  # Sets layer thicknesses when 'double depth' is active
             mask_indices = np.where(grid["mask"] == 1)
-            split = grid["split"]
+            split = grid["split"] - 1
             for n, split_start in enumerate(split[:-1]):
-                depth_value = (2.0**n) * grid["max_subZ"]
+                depth_value = (2.0 ** (n + 1)) * grid["max_subZ"]
```
BenjaminRodenberg, line 592: Needed here? Otherwise better factor this out.

If this indexing shift is needed for the performance optimization you could also apply this change before you merge this PR.

src/ebfm/core/LOOP_climate_forcing.py
```diff
@@ -117,7 +117,7 @@ def set_random_weather_data(IN, C, time, grid):
 
     # Air temperature (K)
     T_amplitude = 10.0  # Seasonal temperature amplitude (K)
-    T_mean_sea_level = 268.0  # Mean sea level temperature (K)
+    T_mean_sea_level = 269.0  # Mean sea level temperature (K)
```
BenjaminRodenberg, line 123: Needed here? Otherwise better factor this out.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -3,9 +3,450 @@
...
+            return fn
+
+        # Handle both @njit and @njit(parallel=True, ...) call styles
+        return _wrap if kwargs or (args and not callable(args[0])) else args[0]
+
+    prange = range  # type: ignore[assignment]  # noqa: F811
+
+# Dispatch flag, false by default (opt-in via --with-numba).
+# Set to True by main.py when --with-numba is passed and numba is properly available.
+_USE_NUMBA = False
```
BenjaminRodenberg: Looks a bit like this should be documented in the argument parser. I'm just thinking about the future situation where the name of the option changes and we forget to update it here.

In the end you are in a different layer of the code here.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -3,9 +3,450 @@
...
+
+
+# Kernels for compaction, heatconduction and percolation steps
+# Compiled with numba when available and _USE_NUMBA is True.
```
BenjaminRodenberg: Maybe put the kernels into a new file (`LOOP_SNOW_kernels.py` or `core/LOOP_SNOW/kernels.py` or `core/kernels/LOOP_SNOW.py`)? But maybe what you are doing here is also perfectly fine and what I'm suggesting goes against existing standards. Just an idea...

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -206,6 +648,61 @@ def compaction():
...
+                C["rd"],
+                C["Ec"],
+                C["Eg"],
+                C["dayseconds"],
+                48 * 2 * SECONDS_PER_HOUR,
+                _mode,
+            )
+            return True
+
+        # Original NumPy path:
```
BenjaminRodenberg: I would rather put this block into an `else:` or add `assert(not _USE_NUMBA)`. Early returning (like the `return True` above) is quite common to reduce indentation, but I think more in an "exception treatment"-sense. Example:

```
if err:
    return fallback_value

value = do_normal_computation(...)
return value
```

However, the situation here is a bit different since there are two exclusive paths. I would prefer

```
if do_numba:
   return do_numba_computation(...)
else:  # no numba
   return do_normal_computation(...)
```

or 

```
if do_numba:
    result = do_numba_computation(...)
else:  # no numba
    result = do_normal_computation(...)

return result
```

In YAC, we tend to use the second option because you only have a single return and not the risk to oversee additional return paths.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -277,36 +775,34 @@ def compaction():
...
 
             z_i = np.zeros_like(OUT["subZ"])
             if nl > 1:
                 z_i[:, 1:] = np.cumsum(OUT["subZ"][:, :-1] * (3.25 - SI[:, :-1]), axis=1)
             gamma_drift = np.maximum(0, SI * np.exp(-z_i / 0.1))
             tau = 48 * 2 * SECONDS_PER_HOUR
-            np.seterr(divide="ignore")
-            tau_i = tau / gamma_drift
+            with np.errstate(divide="ignore", invalid="ignore"):
+                tau_i = tau / gamma_drift
```
BenjaminRodenberg: I don't know what this does. I assume you do not need a comment here if the reader knows that the code does. Could you elaborate briefly here in the review?

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -333,6 +829,7 @@ def compaction():
 
         return True
 
+    @profile
```
BenjaminRodenberg, line 409: Is it also possible to turn off profiling?

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -343,59 +840,86 @@ def heat_conduction():
...
         # Stability time step (CFL condition)
-        z_temp = OUT["subZ"][:, 1:]
-        c_eff_temp = c_eff[:, 1:]
-        kk_temp = kk[:, 1:]
         dt_stab = (
-            0.5 * np.min(c_eff_temp, axis=1) * np.min(z_temp, axis=1) ** 2 / np.max(kk_temp, axis=1) / C["dayseconds"]
+            0.5
+            * np.min(c_eff[:, 1:], axis=1)
+            * np.min(OUT["subZ"][:, 1:], axis=1) ** 2
+            / np.max(kk[:, 1:], axis=1)
```
BenjaminRodenberg, line 424: Is is correct to start with the index `1:` here? Why exclude `OUT["subZ"][:, 0]` etc.?

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -343,59 +840,86 @@ def heat_conduction():
...
-            # Calculate vertical heat fluxes
-            kdTdz[cond_dt, 1] = (
-                (kk[cond_dt, 0] * OUT["subZ"][cond_dt, 0] + 0.5 * kk[cond_dt, 1] * OUT["subZ"][cond_dt, 1])
-                * (subT_old[cond_dt, 1] - OUT["Tsurf"][cond_dt])
-                / dz1[cond_dt]
-            )
+        # subZ and c_eff do not change
+        # Precompute kk*subZ products once
+        kk_sz_top = kk[:, 0] * OUT["subZ"][:, 0] + 0.5 * kk[:, 1] * OUT["subZ"][:, 1]
+        kk_sz_mid = kk[:, 1:-1] * OUT["subZ"][:, 1:-1] + kk[:, 2:] * OUT["subZ"][:, 2:]
```
BenjaminRodenberg, line 431:
```suggestion
        kk_sz_mid = kk[:, 1:-1] * OUT["subZ"][:, 1:] + kk[:, 2:] * OUT["subZ"][:, 2:]
```
The `-1` should not be needed.

If this is similar to the situation with `OUT["subZ"][:, 0]` above, I would actually prefer to use a similar structure as you are using here. The `kk_sz_top` / `kk_sz_mid` naming makes the situation quite clear. (even though `_mid` is an odd name because it's not only the middle layers but all layers except the top layer?)

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -343,59 +840,86 @@ def heat_conduction():
...
 
-            kdTdz[cond_dt, 2:] = (
-                (kk[cond_dt, 1:-1] * OUT["subZ"][cond_dt, 1:-1] + kk[cond_dt, 2:] * OUT["subZ"][cond_dt, 2:])
-                * (subT_old[cond_dt, 2:] - subT_old[cond_dt, 1:-1])
-                / dz2[cond_dt]
-            )
+        # Precompute full temperature-update denominators once
+        denom_l1 = c_eff[:, 1] * (0.5 * OUT["subZ"][:, 0] + 0.5 * OUT["subZ"][:, 1] + 0.25 * OUT["subZ"][:, 2])
```
BenjaminRodenberg, line 434: What is `_l1`? `_layer1`? Is it `_layer1` or actually `_layer0`? To not get too deep into indexing confusion: Maybe again use something like `top` (if it is the top).

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -343,59 +840,86 @@ def heat_conduction():
...
 
-            OUT["subT"][cond_dt, 2:-1] = subT_old[cond_dt, 2:-1] + C_day_dt[:, np.newaxis] * (
-                kdTdz[cond_dt, 3:] - kdTdz[cond_dt, 2:-1]
-            ) / (
-                c_eff[cond_dt, 2:-1]
-                * (
-                    0.25 * OUT["subZ"][cond_dt, 1:-2]
-                    + 0.5 * OUT["subZ"][cond_dt, 2:-1]
-                    + 0.25 * OUT["subZ"][cond_dt, 3:]
+        else:
```
BenjaminRodenberg, line 459: See comment on `if...else` above. I think the structure here is much better.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -414,13 +938,74 @@ def heat_conduction():
...
+            OUT["runoff_surface"] = _runoff_surface
+            OUT["runoff_slush"] = _runoff_slush
+            OUT["refr_P"] = _refr_P
+            OUT["refr_S"] = _refr_S
+            OUT["refr_I"] = _refr_I
+            OUT["refr"] = _refr_P + _refr_S + _refr_I
+            OUT["slushw"] = _slushw
+            OUT["irrw"] = _irrw
+            OUT["cpi"] = 152.2 + 7.122 * OUT["subT"]
+            return True
```
BenjaminRodenberg: See comment on `if...else` above.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -597,71 +1186,113 @@ def layer_merging_and_splitting():
...
+                OUT["subT"][idx_split, split:] = subT_old[idx_split, split - 1 : -1]
+                OUT["subD"][idx_split, split:] = subD_old[idx_split, split - 1 : -1]
 
                 # Update runoff contributions
-                OUT["runoff_irr_deep"][cond_split] += subW_old[cond_split, -1]
-                OUT["runoff_slush"][cond_split] += subS_old[cond_split, -1]
+                OUT["runoff_irr_deep"][idx_split] += subW_old[idx_split, -1]
+                OUT["runoff_slush"][idx_split] += subS_old[idx_split, -1]
 
         return True
```
BenjaminRodenberg, line 877: What does `True` mean here (and also at other places)? Maybe use a constant `SUCCESS = True` here to improve readability.

src/ebfm/main.py
```diff
@@ -250,6 +307,62 @@ def main():
         help="If provided, log output will be written to the specified file (one file per MPI rank).",
     )
 
+    diag_group = parser.add_argument_group("diagnostics, reference snapshots, random seed")
```
BenjaminRodenberg, line 350:
```suggestion
    diagnostics_group = parser.add_argument_group("diagnostics, reference snapshots, random seed")
```

`diag` could also be misunderstood as "diagonal" (which does not make any sense but requires some thinking effort on the reader's side)

src/ebfm/main.py
```diff
@@ -276,6 +389,51 @@ def main():
     logger = getLogger(__name__)
     logger.info(f"Starting EBFM version {ebfm.core.get_version()}...")
 
+    # Numba is opt-in: only activate when --with-numba is explicitly passed.
```
BenjaminRodenberg, line 452:
```suggestion
```

Really needed? Somewhat clear from the parser.

src/ebfm/main.py
```diff
@@ -276,6 +389,51 @@ def main():
...
+    # set_num_threads() must be called before any kernel runs.
```
BenjaminRodenberg: Possible to check this condition with an assertion? Would be much safer than writing a comment here. I imagine something like

```
assert(get_num_threads() != UNSET)  # assuming `get_num_threads` returns `UNSET = -1` if unset.
```

src/ebfm/main.py
```diff
@@ -276,6 +389,51 @@ def main():
...
+            )
+        else:
+            if n_threads < 1:
+                parser.error("--numba-threads must be a >= 1")
+            if n_threads > suggested_max_threads:
+                parser.error(
+                    f"[NUMBA] --numba-threads={n_threads} exceeds suggested max {suggested_max_threads} "
+                    f"(cpu_count={cpu_count}, MPI world size={mpi_size}). "
+                    f"--numba-threads must be between 1 and {suggested_max_threads} on this system. "
+                )
```
BenjaminRodenberg: isn't this the check I'm suggesting above? If it is checked here I think you do not really need the comment. Maybe consider checking such conditions at the very beginning if possible.

src/ebfm/main.py
```diff
@@ -276,6 +389,51 @@ def main():
...
+
+            logger.info(
+                f"[NUMBA] --with-numba enabled. Threads per rank: {n_threads} "
+                f"(suggested max {suggested_max_threads}; cpu_count={cpu_count}, MPI world size={mpi_size})."
+            )
+
+        _numba_mod.set_num_threads(n_threads)
+        LOOP_SNOW._USE_NUMBA = True
```
BenjaminRodenberg: Now I think I get the workflow here. Maybe put the whole `if args.with_numba:...` block into a function? I got the impression that something else than `set_num_threads` happens here but actually this is the whole point of this block. So why not implement this in a small utility function?

```
if args.with_numba:
    set_numba_num_threads(...)
```

or

```
if args.with_numba:
    init_numba(..., args)
```

src/ebfm/main.py
```diff
@@ -302,6 +460,10 @@ def main():
     time = time_config.to_dict()
     grid, io, phys = INIT.init_config(time_config, grid_config, args.restart_dir, args.restart_init)
 
+    if args.random_seed is not None:
+        np.random.seed(args.random_seed)
+        logger.info(f"[DIAG] NumPy random seed fixed to {args.random_seed} for reproducibility.")
```
BenjaminRodenberg, line 494:
```suggestion
        logger.info(f"NumPy random seed fixed to {args.random_seed} for reproducibility.")
```

If you think you need a prefix here this indicates that you could also move the functionality into a different module. Then the logger will automatically prefix the message with the name of the respective file.

You could, for example, do

```
import ebfm.diagnostics

if args.random_seed is not None:
    diagnostics.set_random_seed(args.random_seed)
```

and implement the actual functionality in `ebfm.diagnostics`

src/ebfm/main.py
```diff
@@ -371,6 +532,9 @@ def main():
         # Calculate surface mass balance
         OUT = LOOP_mass_balance.main(OUT, IN, C)
 
+        if args.diagnostics:
+            print_diagnostics(logger, grid, OUT, t)
```
BenjaminRodenberg, line 563:
```suggestion
        if args.print_diagnostics:
            print_diagnostics(logger, grid, OUT, t)
```
?

if the argument only enables printing I would make this more explicit. Currently, one could misunderstand this that everything w.r.t diagnostics is disabled.

src/ebfm/main.py
```diff
@@ -120,6 +121,62 @@ def extract_active_coupling_features(args: argparse.Namespace) -> list[str]:
...
+    """Log key diagnostic values each timestep for performance and correctness analysis."""
+
+    gpsum = grid.get("gpsum", "N/A")
+    has_shading = grid.get("has_shading", False)
+    smb = OUT.get("smb")
+    smb_cum = OUT.get("smb_cumulative")
+
+    logger.info(f"[DIAG t={t + 1}] gpsum={gpsum}, shading={'on' if has_shading else 'off'}")
+    if smb is not None:
+        logger.info(f"[DIAG t={t + 1}] smb:            min={smb.min():.4e}  max={smb.max():.4e}  mean={smb.mean():.4e}")
```
BenjaminRodenberg, line 170: I think now I see the reason for introducing the `[DIAG]` label. I believe you can introduce an independent logger for this purpose. This allows you also to apply formatting rules (like adding this label) to just that logger. This would allow you to call `diagnostics_logger.debug(...)` for diagnostics stuff and `logger.debug(...)` in all other situations.

BenjaminRodenberg: Maybe there is also a nice way to avoid the need to always write `DIAG t={t+1}` if you introduce a dedicated logger. But I'm not sure...

src/ebfm/main.py
```diff
@@ -120,6 +121,62 @@ def extract_active_coupling_features(args: argparse.Namespace) -> list[str]:
...
+    if smb_cum is not None:
+        logger.info(
+            f"[DIAG t={t + 1}] smb_cumulative: "
+            f"min={smb_cum.min():.4e}  "
+            f"max={smb_cum.max():.4e}  "
+            f"mean={smb_cum.mean():.4e}"
+        )
```
BenjaminRodenberg, line 177: You are duplicating some code here. Maybe use a small helper function? `print_smb_stats(smb)/print_smb_stats(smb_cum)` that returns the string `min=... max=... mean=...` with the formatting you want?

BenjaminRodenberg (commented):
Thanks for all the work 🙏 I think there is also already some significant speedup (did not test it but I heard rumors)

If you prefer merging this first and revisiting the PR style-wise that's also fine from my perspective.

src/ebfm/core/LOOP_SNOW.py
BenjaminRodenberg: I got confused with the syntax. `OUT["subZ"][:, 1:-1]` excludes the last element and `OUT["subZ"][:, 1:]` includes the last element. So what I'm suggesting is partially wrong.

Feel free to ignore this comment.

src/ebfm/core/INIT.py
matthias-ibach: Yes, it makes sense to factor this bugfix out. This was done in #116.

src/ebfm/core/LOOP_climate_forcing.py
matthias-ibach: See comment above, addressed in #116.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: Good comment! The `@profile` decorator is actually a no-op at runtime. 
There is a `try/except` fallback at the top of the file to catch that. \
So when not running under `kernprof`, profile is just an identity function with zero overhead. No code change needed. It is intentionally left in so you can profile by simply swapping `python` for `kernprof -l` without touching the source code. This can, however, be improved in the future!

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: See comment from your first statement.

src/ebfm/main.py
matthias-ibach: Was removed.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: Great point! This has been fully refactored. \
The `_USE_NUMBA` flag and its comment are gone. I introduced a `compute_backend.py` with a `ComputeBackend` `enum` and an `init_numba(n_threads)` function. 

`main.py` calls `init_numba()` after CLI parsing. Now that is the single place that activates the backend. The comment coupling the flag to `--with-numba` is no longer needed since activation goes through a proper API call.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: Done! The three kernels (`_compaction_kernel, _heat_conduction_kernel, _percolation_kernel`) have been extracted into `LOOP_SNOW_kernels.py` and are simply imported in `LOOP_SNOW.py`. I chose a flat `core/LOOP_SNOW_kernels.py` file to stay consistent with the existing module layout.
This can also be changed in a future MR, if that might be better.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: This has been addressed. Each dispatch block now uses your suggested pattern:

```
if get_backend() == ComputeBackend.NUMBA:
    ...
else:
    ...
return True
```
There is a single return at the bottom, making both paths visually symmetric and explicit.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: See comment above , has been resolved.

src/ebfm/main.py
matthias-ibach: This has already been implemented. `init_numba()` in `compute_backend.py` calls `set_num_threads()` and immediately asserts `get_num_threads() >= 1` to confirm it was applied.

src/ebfm/main.py
matthias-ibach: Agreed! The validation is now in the dedicated `_compute_numba_threads()` helper in `main.py`, which runs all checks (range, suggested max) before `init_numba()` is called. The comment has been removed since the code itself expresses the constraint.

src/ebfm/main.py
matthias-ibach: Done! The entire `if args.with_numba` block is now split into two clean calls:

```
n_threads = _compute_numba_threads(args, ebfm_comm, parser, logger)
init_numba(n_threads)
``` 

`_compute_numba_threads()` handles validation and logging and `init_numba()` in `compute_backend.py` handles thread setup and backend activation.

PR #106, opened 2026-04-08
Fix units for SMB being passed to Elmer/Ice and develop general functionality for unit conversion

ClemensSchannwell:
At the moment, we are passing the variable `out['SMB']` via YAC to Elmer. However, the units are in `meters per timestep`. This is not what we want as Elmer works with `meters per year`. This PR fixes this and should work for arbitrary time steps in EBFM. 

# Author's Todos

### DIFF
--- a/src/ebfm/coupling/components/icon_atmo.py
+++ b/src/ebfm/coupling/components/icon_atmo.py
@@ -117,16 +117,21 @@ class IconAtmo(Component):
         """
         received_data: dict[str, np.ndarray] = {}
 
+        # We need to convert precipitation received from ICON from kg / m^2 / s
+        # to the unit expected by EBFM (???)
+        def map_pr_icon_to_ebfm(x: np.ndarray) -> np.ndarray:
+            seconds_per_year = DAYS_PER_YEAR * SECONDS_PER_DAY
+            return x * seconds_per_year * 1e-3
+
         # Put data to IconAtmo
         self._put_if_coupled("albedo", data_to_exchange)
 
         # Get data from IconAtmo
-        pr = self._get_if_coupled("pr")
+        pr = self._get_if_coupled("pr", transform=map_pr_icon_to_ebfm)
         if pr is not None:
-            # convert precipitation from kg m-2 s-1 to m w.e.
-            received_data["pr"] = pr * self._coupler.get_seconds_per_year() * 1e-3
+            received_data["pr"] = pr
 
-        # TO DO: check what units are needed for pr_snow in EBFM
+        # TODO: check what units are needed for pr_snow in EBFM
         pr_snow = self._get_if_coupled("pr_snow")
         if pr_snow is not None:
             received_data["pr_snow"] = pr_snow
--- a/src/ebfm/coupling/components/icon_atmo.py
+++ b/src/ebfm/coupling/components/icon_atmo.py
@@ -10,7 +10,8 @@ if TYPE_CHECKING:
 
 from .base import Component
 
-from ebfm.coupling.fields import FieldSet, Field, ExchangeType, days_to_iso
+from ebfm.coupling.fields import FieldSet, Field, ExchangeType, Timestep
+from ebfm.core.config import TimeConfig
 
 
 class IconAtmo(Component):
@@ -23,11 +24,11 @@ class IconAtmo(Component):
     def __init__(self, coupler: "Coupler"):
         super().__init__(coupler)
 
-    def get_field_definitions(self, time: dict[str, float]) -> FieldSet:
+    def get_field_definitions(self, time: TimeConfig) -> FieldSet:
         """
         Get generic field definitions for EBFM coupling to IconAtmo.
         """
-        timestep = days_to_iso(time["dt"])
+        timestep = Timestep(value=time.time_step_iso8601())
 
         return FieldSet(
             {
--- a/src/ebfm/coupling/components/icon_atmo.py
+++ b/src/ebfm/coupling/components/icon_atmo.py
@@ -5,7 +5,7 @@
 from typing import TYPE_CHECKING
 from collections.abc import Mapping
 import numpy as np
-from ebfm.core.constants import DAYS_PER_YEAR, SECONDS_PER_DAY
+from ebfm.core.constants import SECONDS_PER_DAY
 
 if TYPE_CHECKING:
     from ebfm.coupling.couplers.base import Coupler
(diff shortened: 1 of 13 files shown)


BenjaminRodenberg:
Where to perform conversion of units is a tricky design question (therefore adding the `generic coupling interface` label) and there are a few different solutions:

* Sender converts
* Receiver converts
* Coupler converts (I'm not aware if YAC can actually do this)

You are picking "Sender converts" here. Nothing wrong about this but I would rather see such a scaling of units in the coupling module of EBFM instead of in the main file. If you, for example, swap Elmer for a different glacier model I think it would be ok to touch the code in the coupling module but the EBFM model itself should not be affected by such a change.

Maybe one more question: Why not let Elmer take care of the conversion?

BenjaminRodenberg:
I added the `bug` label because currently the units do not match. As a quick bugfix this PR is fine. Coming up with a general system for more comfortable unit conversion (maybe even without the necessity to change the code but through some configuration at runtime) is a bigger feature and something that could be solved as a follow-up if needed.

I would still try to shift the code changes to a better place as suggested above.

ClemensSchannwell:
I am happy to move the code and it would even be more convenient to specify the units outside of the source code.  
 
> Maybe one more question: Why not let Elmer take care of the conversion?

Hmm, in my mind every model component should send data in some SI units or at least commonly used units. For me meters/timestep is not something I would ever expect. This needs knowledge about the model that I do not expect standard users to have. Especially, because this could be different from the coupling timestep. I still do the conversion to ice-units within Elmer, but I think this should be done in the dedicated component model which in this case is Elmer/Ice. By the way, the same issue arises for runoff, which we also need in m/year instead of meters/timestep

ClemensSchannwell:
@BenjaminRodenberg  I moved the implementation to the coupling class now. Can you have a look if it is now better implemented?

src/ebfm/coupling/couplers/base.py
```diff
@@ -98,9 +101,52 @@ def get_component(self, component_name: str) -> Component:
...
+
+        @note This method assumes that self._time has been set (i.e. self.setup() has been called).
+
+        @return The conversion factor (unitless).
+        """
+        assert self._time is not None, "self._time must be set before calling get_conversion_per_year_factor."
+
+        ebfm_time_step = self._time.time_step_in_days()
+
+        return DAYS_PER_YEAR / ebfm_time_step
```
BenjaminRodenberg: This could be shifted inside `TimeConfig`. But if it is more convenient the way it is right now. I can also live with this solution.

src/ebfm/main.py
```diff
@@ -335,9 +335,9 @@ def main():
             logger.debug("Done.")
             logger.debug("Received the following data from ICON:", data_from_icon)
 
-            IN["P"] = (
-                data_from_icon["pr"] * time["dt"] * C["dayseconds"] * 1e-3
-            )  # convert units from kg m-2 s-1 to m w.e.
+            # convert precipitation from kg m-2 s-1 to m w.e.
+            # TODO: unit conversion to EBFM local units should happen inside the coupler
+            IN["P"] = data_from_icon["pr"] * time_config.time_step.total_seconds() * 1e-3
```
BenjaminRodenberg: Should be moved inside the ICON coupling. I think you can test this much better with your setup. But we can also do this later.

BenjaminRodenberg (changes requested):
I reworked a part of this PR to consequently use the `TimeConfig` object in the `Coupler`. One easily messes up unit conversions and, therefore, I'm trying to offer small helpers in `TimeConfig`.

The `to_dict()` function still allows to provide the legacy dictionary to parts of the code I do not want to touch (yet).

@ClemensSchannwell Do you want to take a look at the highlighted parts? I do not think it is a must, but since you are already working with this part now I think it might be a good idea.

src/ebfm/coupling/couplers/base.py
ClemensSchannwell: I think I would leave it where it is right now. But when it comes to design questions, I do not think I should be the measuring stick.

src/ebfm/main.py
ClemensSchannwell: Now that we are working on this already, I think it would be good to know @wardvp if snow also needs to be in units m w.e. or something else?

ClemensSchannwell: I also think that the conversion is wrong if EBFM really needs it as m w.e. because we would have to convert `data_from_icon["pr"] * SECONDS_PER_YEAR * 1e-3`

src/ebfm/coupling/components/icon_atmo.py
```diff
@@ -104,24 +107,31 @@ def get_field_definitions(self, time: dict[str, float]) -> FieldSet:
...
 
         @returns dictionary of received field data
         """
         received_data: dict[str, np.ndarray] = {}
 
+        # We need to convert precipitation received from ICON from kg / m^2 / s
+        # to the unit expected by EBFM (???)
+        def map_pr_icon_to_ebfm(x: np.ndarray) -> np.ndarray:
+            seconds_per_year = DAYS_PER_YEAR * SECONDS_PER_DAY
+            return x * seconds_per_year * 1e-3
```
BenjaminRodenberg: @ClemensSchannwell I moved the transformation into a function that is automatically applied to the returned data.

The main motivation for using a function here is that this can be reused more easily elsewhere. It also gives us the chance to add a bit more information on what's actually happening with the name of the function and the docstring.

Can you maybe help me here with the units? As far as I understand ICON gives us kg / m^2 / s. This function converts it to m / year ? Is this a reasonable unit or would you phrase this differently? (Feel free to just edit)

src/ebfm/coupling/couplers/base.py
BenjaminRodenberg: It is definitely necessary to store the time config in the Coupler. Whether the Coupler provides the time step size (in any unit) in it's original form or if we first apply some conversion is in the end a detail. I think for now this place is ok. If at some point we need to convert more units and a different place looks more natural we can still move it.

src/ebfm/main.py
BenjaminRodenberg: I see. You actually changed the factor from `pr_from_icon * time["dt"] * C["dayseconds"] * 1e-3` to `pr_from_icon * DAYS_PER_YEAR * SECONDS_PER_DAY * 1e-3`.

I cannot judge what's correct but it's important to clarify this. I would try to solve this independently: If `pr_from_icon * time["dt"] * C["dayseconds"] * 1e-3` is wrong that`s a bug and we should fix it in an independent PR. This also makes it a bit easier to follow the changes here.

We can of course discuss this issue here first and when we have clarify I can clean things up before merging.

ClemensSchannwell: I understand and agree with you that we shouldn't change units here, but open a separate PR for this. Having said this, I think in my enthusiasm I misread the units because I think it said `m w.e.` instead of what I read `m w.e./year`. But we should check with @wardvp what the correct units in EBFM for `precip` and `snowfall` are.

src/ebfm/coupling/components/icon_atmo.py
ClemensSchannwell: This is what the function does, but I think we should go back to the original implementation: `pr * time["dt"] * SECONDS_PER_DAY * 1e-3`. This converts  kg / (m^2 s) to meters of water equivalent (m w.e.). 
Can you tell me why you make a separate variable `seconds per day` in this function?

BenjaminRodenberg: I was always wondering what "m w.e." means. Thanks for the explanation :+1: 

I'm introducing `seconds_per_year` only for readability reasons. This saves a (tiny amount) of additional thinking.

BenjaminRodenberg: I'll then take care of rolling this back to the original implementation.

BenjaminRodenberg (approved):
From my perspective this PR would be good to merge now.

We still need to clarify some details about conversion of other units but as I understood @ClemensSchannwell we will do this in a follow-up issue / PR.

ClemensSchannwell:
Lets merge!

PR #109, opened 2026-04-29
Add logger info coupling components

matthias-ibach:
Add logger information to base class of coupling components. Add metadata to smb ("Surface mass balance").

# Author's Todos

### DIFF
--- a/src/ebfm/coupling/components/base.py
+++ b/src/ebfm/coupling/components/base.py
@@ -7,6 +7,10 @@ from typing import TYPE_CHECKING
 from collections.abc import Mapping, Callable
 import numpy as np
 
+from ebfm.core import logging
+
+logger = logging.getLogger(__name__)
+
 if TYPE_CHECKING:
     from ebfm.coupling.couplers.base import Coupler
     from ebfm.coupling.fields.base import FieldSet
@@ -66,7 +70,10 @@ class Component(ABC):
             assert (
                 field_name in data_to_exchange
             ), f"Field '{field_name}' is missing in data_to_exchange for component '{self.name}'."
-            self._coupler.put(self.name, field_name, transform(data_to_exchange[field_name]))
+            logger.debug(f"Putting field {field_name=}")
+            err = self._coupler.put(self.name, field_name, transform(data_to_exchange[field_name]))
+            if err:
+                logger.warning(f"Error code {err=} after put of {field_name=}")
 
     def _get_if_coupled(self, field_name: str, transform: Callable = identity) -> np.ndarray | None:
         """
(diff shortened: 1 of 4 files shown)


src/ebfm/coupling/components/base.py
```diff
@@ -66,7 +70,10 @@ def _put_if_coupled(
             assert (
                 field_name in data_to_exchange
             ), f"Field '{field_name}' is missing in data_to_exchange for component '{self.name}'."
-            self._coupler.put(self.name, field_name, transform(data_to_exchange[field_name]))
+            logger.debug(f"Putting field {field_name=}")
+            err = self._coupler.put(self.name, field_name, transform(data_to_exchange[field_name]))
+            if err:
+                logger.warning(f"Error code {err=} after put of {field_name=}")
```
BenjaminRodenberg:
```suggestion
            if err:
                logger.warning(f"Put for {field_name=} returned error code ({err=})")
```
Maybe this rephrasing helps to tell the user that something went wrong and the error code is just some additional information the user does not necessarily have to worry about.

BenjaminRodenberg (changes requested):
Raising a warning instead of just ignoring the error code is certainly a good direction. 

What I currently find a bit odd is to provide an error code to the user without telling the user what to do with it (I would expect such information on debug level when the user asks for it and not so much as a warning). If you want, you could catch specific error codes and provide guidance on how to fix the situation. If this is too complicated, maybe add a message that helps the user to decipher the error code:

* Refer to docs / troubleshooting for a list of error codes and their meaning
* Ask the user to report to developers
* Rephrase as shown in comment

Additionally, I think it actually would not harm to briefly mention this in the changelog as well because this is a user-observable change. This also gives you a place to explain the rationale behind this and why it is a warning and not an error.

src/ebfm/coupling/components/base.py
matthias-ibach: This suggestion has been applied in the code as:
```
            if err:
                logger.warning(
                    f"Put for {field_name=} returned error code ({err=}). Please report this to the developers."
                )
```

matthias-ibach (commented):
Thank you for the valuable comment.
In general, catching specific error codes and mapping them to actionable guidance makes absolutely sense, but it would require deeper knowledge of each coupler's error space. This might be out of scope for this PR. 

What I have done in this PR is adding a changelog entry and applying the rephrasing suggestion. This should make it clear that the error code is supplementary info. 

We could revisit a more detailed error handling in a separate PR or by addressing this in a follow-up issue, if needed!

BenjaminRodenberg (approved):

PR #110, opened 2026-04-30
Make DoF location for exchanged fields configurable

BenjaminRodenberg:
I changed the DoF location from cell corners to centers in #83. The motivation here was that the exchanged fields are conservative fields which need to be defined per-cell.

I now have a different use case where I just want to receive a field from another component and then I want to plot this field. The easiest way to do so is to just receive data on vertices - cell data would require additional connectivity information.

Even though my use-case is only a prototype development I still think it would be useful to be able to configure whether a field should be received on vertices or corners.

The development of a generic interface that allows configuration on component level is still a todo. Alternatively, the coupler could also return data for a fixed pre-defined set of locations, i.e. YAC would compute data for all fields on corners and cell centers by default. This would be a simple approach. However, it would require the user to look inside the specific coupler - thus make the interface less generic.

# Author's Todos

- [ ] I ran the [pre-commit hooks](https://github.com/EBFMorg/EBFM?tab=readme-ov-file#developer-notes) and fixed all issues
- [ ] I added an entry under *develop* in the [CHANGELOG.md](https://github.com/EBFMorg/EBFM/blob/main/CHANGELOG.md) (may be skipped for minor changes not observable for the user)

### DIFF
--- a/src/ebfm/coupling/couplers/yacCoupler.py
+++ b/src/ebfm/coupling/couplers/yacCoupler.py
@@ -268,8 +268,9 @@ class YACCoupler(Coupler[yac.ExchangeType]):
             ), f"Cannot add field '{field.name}' for uncoupled component '{field.coupled_component.name}'."
 
             # TODO: make this configurable in Field base class
-            # dof_location = self.cell_centers
-            dof_location = self.cell_corners
+            # dof_location = self.cell_corners  # breaking existing code
+            
+            dof_location = self.cell_centers  # original setting
 
             yac_field = YACField.from_field(field).construct_yac_field(
                 self.interface, self.component, collection_size, dof_location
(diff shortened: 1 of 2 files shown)


src/ebfm/coupling/couplers/yacCoupler.py
```diff
@@ -264,9 +267,14 @@ def _construct_coupling_pre_sync(self, field_definitions: FieldSet):
                 field.coupled_component.name
             ), f"Cannot add field '{field.name}' for uncoupled component '{field.coupled_component.name}'."
 
+            # TODO: make this configurable in Field base class
+            # dof_location = self.cell_centers
+            dof_location = self.cell_corners
```
BenjaminRodenberg:
```suggestion
            # TODO: make this configurable in Field base class
            # dof_location = self.cell_corners  # breaking existing code
            
            dof_location = self.cell_centers  # original setting
```

PR #111, opened 2026-04-30
Revise component config

BenjaminRodenberg:
While adding a new component I realized that during parsing/configuration there are some places where one can easily forget adding the required functions. I revised the registration of new components in the configuration.

Instead of individual members we now have a dictionary containing `True/False`. If a component is not registered in the dictionary the respective getter should fail (instead of always returning `False`).

Since I wanted to avoid duplication of labels for identifying components I had to move `Component.name` to a different place. Otherwise this would have lead to a circular include. I'm not 100% happy about the fact that a `Component` does not identify it's own name anymore. Instead, the name is defined from the outside. However, this also opens up a new use-case that we could have two different instances of, for example, an `ElmerIce` component which live under different names.

# Author's Todos

### DIFF
--- a/src/ebfm/core/LOOP_EBM.py
+++ b/src/ebfm/core/LOOP_EBM.py
@@ -16,6 +16,10 @@ from ebfm.core import LOOP_EBM_SWout, LOOP_EBM_insolation
 
 from ebfm.coupling import Coupler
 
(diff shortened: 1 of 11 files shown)


k202077 (approved):
based on my limited understanding of Python, it looks fine to me

PR #116, opened 2026-06-01
Bug fixes in double depth method by Ward

matthias-ibach:
This PR addresses the bug fixes in the double depth method introduced by Ward.
These fixes were initially added to #105 and have been extracted here to push them to main in a cleaner, isolated way.

**Note:** The fixes were isolated and applied on top of the current main branch. Because the code in main differs from the heavily refactored code in #105, some of the changes look different, but the logic should be identical.

@wardvp Please feel free to review the isolated bug fixes. If everything looks good, I would suggest merging this first, so the fixes land in main cleanly.

After that, #105 can be merged to bring in the performance optimizations.

@BenjaminRodenberg Let me know if there is anything I missed or if you have any comments.

# Author's Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Bug fixes in double depth method in INIT.py and LOOP_SNOW.py
 * EBFM now only adds metadata to fields where this is explicitly specified. Note: This can lead to failures in components that do not properly guard `get_metadata` calls with `has_metadata` checks. https://github.com/EBFMorg/EBFM/pull/102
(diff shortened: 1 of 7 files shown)


BenjaminRodenberg (approved):
Code generally looks ok to me. Cannot really judge on what you are fixing here but I assume you and Ward know very well what was wrong before and is now fixed.

If possible: Try to add a test that would fail for the buggy version of the code and which is successful for this version including the fix. The main reason here is to avoid running into the same bug again. It's usually much easier to implement such a test now than later because you already have the failing case at hand (otherwise you would not have been able to detect the bug).

wardvp (approved):

PR #117, opened 2026-06-03
Fix: CouplingConfig.comms initialisation causes RuntimeError on startup

matthias-ibach:
Problem: `self.comms = {}` led to a runtime error raising `RuntimeError: Group communicators have already been set.` 

This error is not shown as a failed test, see for example [here](https://github.com/EBFMorg/EBFM/actions/runs/26813778392/job/79050238306). Issue #118 was created to address this issue.

Fix: 
- `self.comms = None` + update the annotation to `dict[str, MPI.Comm] | None`
- Add `# type: ignore` on the two lines where `mypy` cannot narrow through method-call guards. This can be discussed and changed if a different solution is preferred. 

# Author's Todos

### DIFF
--- a/src/ebfm/core/comm/mpi_handshake.py
+++ b/src/ebfm/core/comm/mpi_handshake.py
@@ -8,7 +8,6 @@ from mpi4py import MPI
 import numpy as np
 import sys
 
-
 __version__ = "1.0.0"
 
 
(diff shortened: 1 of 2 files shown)


BenjaminRodenberg (approved):
Thanks for the fix 🙏 

I was a bit too quick fixing the mypy stuff yesterday. Strange that the tests do not detect this regression though...

PR #119, opened 2026-06-03
Revise `reader.py`

BenjaminRodenberg:
* Fixes bug when calling `read_dem` is called.
* Revise algorithm used by `reader.py` to determine nearest neighbor points.
* Improve criterion to determine mismatch of points. We now consider points as matching if they do not differ by more than half the meshwith.

# Author's Todos

### DIFF
--- a/src/ebfm/reader.py
+++ b/src/ebfm/reader.py
@@ -168,7 +168,11 @@ def read_dem(dem_file: Path, xs: NDArray[np.float64], ys: NDArray[np.float64]):
         if not is_monotonic:
             raise ValueError("Axis values must be strictly monotonic (strictly increasing or strictly decreasing).")
 
-        axis_for_search = axis_values if is_ascending else axis_values[::-1]
+        if is_descending:
+            # For searchsorted to work correctly, we need to reverse the axis values if they are descending.
+            axis_for_search = axis_values[::-1]
+        else:
+            axis_for_search = axis_values
 
         n_axis = len(axis_for_search)
         insert_pos = np.searchsorted(axis_for_search, query_values)
@@ -188,8 +192,8 @@ def read_dem(dem_file: Path, xs: NDArray[np.float64], ys: NDArray[np.float64]):
         choose_right = right_dist <= left_dist
         nearest_pos = np.where(choose_right, right_pos_safe, left_pos_safe)
 
-        if is_descending:  # if axis is descending, we reversed it for search, so we need to flip the indices back
-            assert not is_ascending
+        if is_descending:
+            # we reversed it for search, so we need to flip the indices back
             nearest_pos = len(axis_values) - 1 - nearest_pos
 
         return nearest_pos.astype(np.int64)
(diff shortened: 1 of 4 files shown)


src/ebfm/reader.py
```diff
@@ -152,28 +156,123 @@ def read_dem(dem_file: Path, xs: NDArray[np.float64], ys: NDArray[np.float64]):
...
+                "tolerance."
+            )
+            logger.warning(
+                f"{np.sum(mismatch_y > normalized_error_tol)} of {len(idx_y)} y-coordinates do not match within "
+                "tolerance."
+            )
+            logger.warning(
+                f"Maximum spacing-normalized mismatch: {np.max(mismatch_x)} in x and {np.max(mismatch_y)} in y."
+            )
+            print_axis_diagnostics("x", x_axis, xs, idx_x, mismatch_x, x_spacing)
```
matthias-ibach: The print diagnostics are only called for `x`, but not for `y`. Maybe:
```suggestion
            print_axis_diagnostics("x", x_axis, xs, idx_x, mismatch_x, x_spacing)
            print_axis_diagnostics("y", y_axis, ys, idx_y, mismatch_y, y_spacing)            
```

How about the spacing-normalized mismatch?

src/ebfm/reader.py
```diff
@@ -152,28 +156,123 @@ def read_dem(dem_file: Path, xs: NDArray[np.float64], ys: NDArray[np.float64]):
...
+        idx_y = nearest_indices(y_axis, ys)
+
+        # Normalize absolute coordinate mismatch by typical axis spacing.
+        # This keeps diagnostics meaningful near x/y ~= 0.
+        x_spacing = np.median(np.abs(np.diff(x_axis)))
+        y_spacing = np.median(np.abs(np.diff(y_axis)))
+        x_spacing = max(x_spacing, np.finfo(np.float64).eps)
+        y_spacing = max(y_spacing, np.finfo(np.float64).eps)
+
+        normalized_error_tol = 0.5  # warn if nearest neighbor is more than 0.5 typical spacings away from query point
```
matthias-ibach, line 215: Is this a typical error tolerance, more than half a cell?
Sounds plausible to me, just wondering.

src/ebfm/reader.py
```diff
@@ -152,28 +156,123 @@ def read_dem(dem_file: Path, xs: NDArray[np.float64], ys: NDArray[np.float64]):
...
+                queries: NDArray[np.float64],
+                matched_indices: NDArray[np.int64],
+                normalized_mismatch: NDArray[np.float64],
+                axis_spacing: float,
+                max_examples: int = 5,
+            ) -> None:
+                diffs = np.diff(axis_values)
+                is_ascending = np.all(diffs >= 0)
+                is_descending = np.all(diffs <= 0)
+                is_monotonic = is_ascending or is_descending
```
matthias-ibach: In line- 164-165, the logic is:

```
        is_ascending = np.all(diffs > 0)
        is_descending = np.all(diffs < 0)
        is_monotonic = is_ascending or is_descending
```

Maybe for here:

```suggestion
                diffs = np.diff(axis_values)
                is_ascending = np.all(diffs > 0)
                is_descending = np.all(diffs < 0)
                is_monotonic = is_ascending or is_descending
```

matthias-ibach (commented):
MR looks good to me, with just some small minor comments and suggestions.

src/ebfm/reader.py
```diff
@@ -152,28 +156,123 @@ def read_dem(dem_file: Path, xs: NDArray[np.float64], ys: NDArray[np.float64]):
...
+        left_pos_safe = np.where(left_valid, left_pos, 0)
+        right_pos_safe = np.where(right_valid, right_pos, n_axis - 1)
+        left_dist = np.where(left_valid, np.abs(axis_for_search[left_pos_safe] - query_values), np.inf)
+        right_dist = np.where(right_valid, np.abs(axis_for_search[right_pos_safe] - query_values), np.inf)
+
+        choose_right = right_dist <= left_dist
+        nearest_pos = np.where(choose_right, right_pos_safe, left_pos_safe)
+
+        if is_descending:  # if axis is descending, we reversed it for search, so we need to flip the indices back
+            assert not is_ascending
```
matthias-ibach: Is this assertion necessary?

BenjaminRodenberg: This makes the (quite obvious) fact that you cannot have both `is_descending` and `is_ascending` at the same time explicit at this point.

Maybe it is generally a bad idea to have `is_descending` and `is_ascending` at the same time and one could instead only do the following:

```py
is_ascending = np.all(diffs > 0)
if not is_ascending:
    assert np.all(diffs < 0)  # is not is_ascending make sure it must be descending
```

So "is it necessary" is up to discussion ;) How do you like the idea of removing `is_descending` (and probably also `is_monotonic`?

src/ebfm/reader.py
BenjaminRodenberg: Good catch :+1: These kind of errors are actually a good argument for a function `is_ascending(axis_values)`. I can also clarify in the docs that we mean strictly monotonically ascending/descending.

src/ebfm/reader.py
BenjaminRodenberg: Sure.

I do not full understand what you mean with this:

> How about the spacing-normalized mismatch?

matthias-ibach: Sorry about the confusion and being imprecise. 
I was not sure whether another `print_axis_diagnostics` info for the logger warning in lines 267-268 (called "Maximum spacing-normalized mismatch:...") are needed.

BenjaminRodenberg: Ah. Now I get it. Thanks for the clarification :)

src/ebfm/reader.py
BenjaminRodenberg: That was the tolerance the AI provided and I think it makes sense.

Previously the tolerance was definitely too strict and therefore the checks were not really helpful.

PR #122, opened 2026-06-04
Fix format in CHANGELOG.md

BenjaminRodenberg:
Minor formatting fixes


matthias-ibach (commented):
Thanks for fixing the formatting issue, this must have slipped through my fingers.

Thank you also for making me aware of this! I will keep this in mind for future changes in the `CHANGELOG.md`.

Otherwise, this small fix should be good to go.

PR #125, opened 2026-06-16
Forward EBFM's start/end time and calendar to YAC

BenjaminRodenberg:
Call `def_datetime` to avoid the following risks:

1) Inconsistency between EBFM's and YAC's time frame
2) Error if starttime/endtime is missing in config

Call `def_calendar` to:

1) explicitly tell YAC about the calendar in use
2) prevent failures due to missing calendar definition
3) Remark: EBFM only supports a Gregorian calendar, but may "lie" to YAC as a workaround. See #127 for details.

# Author's Todos


src/ebfm/coupling/couplers/yacCoupler.py
```diff
@@ -50,6 +50,9 @@ def __init__(self, coupling_config: CouplingConfig):
             logging.debug(f"YAC uses the following comm from EBFM's mpi-handshake: {yac_group_label}: {coupling_comm}.")
             self.interface = yac.YAC(comm=coupling_comm)
 
+        self.interface.def_starttime(coupling_config.start_time)
+        self.interface.def_endtime(coupling_config.end_time)
```
k202077: Are you sure that it is not:
```
self.interface.def_datetime(coupling_config.start_time, coupling_config.end_time)
```
?

BenjaminRodenberg: Sure. Thanks for the hint :pray: 

Unfortunately, the tests do not trigger here since a mock coupler does not know about the actual interface of YAC.

src/ebfm/core/config/time.py
```diff
@@ -0,0 +1,117 @@
...
+        step_seconds = self.time_step.total_seconds()
+        assert total_seconds % step_seconds == 0, "Time interval must be divisible by time step."
+        return int(round(total_seconds / step_seconds))
+
+    def time_step_in_days(self) -> float:
+        """Get the time step size in days.
+
+        @returns Time step size in days
+        """
+        return self.time_step.total_seconds() / SECONDS_PER_DAY
```
k202077, line 94: There are leap seconds! If a simulation includes one of those days....

src/ebfm/coupling/couplers/yacCoupler.py
```diff
@@ -33,6 +35,19 @@ def __init__(self, coupling_config: CouplingConfig):
...
+                yac.def_calendar(yac.Calendar.PROLEPTIC_GREGORIAN)
+            case Calendar.YEAR_OF_365_DAYS:
+                yac.def_calendar(yac.Calendar.YEAR_OF_365_DAYS)
+            case Calendar.YEAR_OF_360_DAYS:
+                yac.def_calendar(yac.Calendar.YEAR_OF_360_DAYS)
+            case _:
+                raise ValueError(
+                    f"Unsupported calendar '{coupling_config.time_config.calendar}'. Supported calendars are: "
+                    f"{[c.value for c in yac.Calendar]}"
+                )
```
BenjaminRodenberg, line 49: > Due to the split of the files, I cannot see the lines that actually changed.

I assume that's the part you refer to?

> Did you try to reproduce the above mentioned setup?

Yes. I played around with the configuration and commented out the definition of the calendar in the config. This directly triggered `ERROR(set_datetime): calendar has not yet been set`. When setting the calendar explicitly via `def_calendar` this also revealed some inconsistencies across components. So I decided to generally define the calendar in the components to allow YAC detecting potential inconsistencies.

src/ebfm/core/config/time.py
BenjaminRodenberg: EBFM only supports very rudimentary time-keeping. I opened an issue to also support 360/365-day calendars (#127) but leap seconds are another important point.

I already offloaded some parts to Python's datetime. But I think this doesn't help here because we are leaving the safe ground of datetime by manually computing `time_step_in_days` as a `float` here.

I would suggest: Let's live with the fact that EBFM is inaccurate in terms of 360/365-day calendars and leap seconds in this PR. But we certainly need to fix it at some point. I'm not an expert here but I assume for relatively short simulation times it only introduces a slight error?

src/ebfm/core/config/time.py
```diff
@@ -0,0 +1,117 @@
...
+
+        self.dT_UTC = 1  # Time difference relative to UTC in hours (hard-coded for now)
+
+        self.calendar = Calendar(args.calendar)
+
+        if self.calendar is not Calendar.PROLEPTIC_GREGORIAN:
+            logger.warning(
+                f"Using calendar {self.calendar.value}. Note that support for non-proleptic-gregorian calendars is "
+                f"experimental and may lead to unexpected behavior."
+            )
```
BenjaminRodenberg, line 77: This warning is my approach to tell the user: "You may ask for a calendar different than the Gregorian one. But technically this has no actual effect and can lead to errors (esp. for longer simulations)."

We could make this a bit more nuanced by checking the total simulation time and return another warning (or even a fatal error?) if ignoring leap seconds or 360/365-day calendars would lead to a significant error. I do not have experience here but I assume there are some general rules?

BenjaminRodenberg (commented):
Added a few comments for clarification.

src/ebfm/core/config/time.py
k202077: or just wait until it crashes for the first time ;-)
Maybe @wardvp has on opinion on that.

BenjaminRodenberg: I discussed this with @ClemensSchannwell and we decided to keep things as they are. I will add more details and future directions in #127.

PR #126, opened 2026-06-17
Improve diagnostics logging info

matthias-ibach:
This merge request addresses open questions that were raised in #105, but not answered there because they did not quite fit the scope at that time and the performance improvement implementations had a higher priority.

The unresolved comments are picked up here again. The comments are pasted at the appropriate locations and the responses (addressing the changes) are written below.

Once this MR is merged, the unresolved issues in #105 will be closed.

# Author's Todos

- [ n/a] I added an entry under *develop* in the [CHANGELOG.md](https://github.com/EBFMorg/EBFM/blob/main/CHANGELOG.md): Changes solely include renaming and restructuring of logging and diagnostics.

### DIFF
--- a/src/ebfm/core/LOOP_SNOW.py
+++ b/src/ebfm/core/LOOP_SNOW.py
@@ -417,6 +417,7 @@ def main(C, OUT, IN, dt, grid, phys):
         c_eff = OUT["subD"] * (152.2 + 7.122 * OUT["subT"])  # Effective heat capacity
 
         # Stability time step (CFL condition)
+        # Layer 0: surface ghost layer, excluded from CFL condition
         dt_stab = (
             0.5
             * np.min(c_eff[:, 1:], axis=1)
@@ -427,11 +428,14 @@ def main(C, OUT, IN, dt, grid, phys):
 
         # subZ and c_eff do not change
         # Precompute kk*subZ products once
+        # kk_sz_top: conductivity-thickness product for the top interface
+        # kk_sz_interior: same for all interior interfaces
         kk_sz_top = kk[:, 0] * OUT["subZ"][:, 0] + 0.5 * kk[:, 1] * OUT["subZ"][:, 1]
-        kk_sz_mid = kk[:, 1:-1] * OUT["subZ"][:, 1:-1] + kk[:, 2:] * OUT["subZ"][:, 2:]
+        kk_sz_interior = kk[:, 1:-1] * OUT["subZ"][:, 1:-1] + kk[:, 2:] * OUT["subZ"][:, 2:]
 
         # Precompute full temperature-update denominators once
-        denom_l1 = c_eff[:, 1] * (0.5 * OUT["subZ"][:, 0] + 0.5 * OUT["subZ"][:, 1] + 0.25 * OUT["subZ"][:, 2])
+        # denom_layer1: first active layer (layer 0: surface ghost layer overwritten from Tsurf)
+        denom_layer1 = c_eff[:, 1] * (0.5 * OUT["subZ"][:, 0] + 0.5 * OUT["subZ"][:, 1] + 0.25 * OUT["subZ"][:, 2])
         denom_interior = c_eff[:, 2:-1] * (
             0.25 * OUT["subZ"][:, 1:-2] + 0.5 * OUT["subZ"][:, 2:-1] + 0.25 * OUT["subZ"][:, 3:]
         )
@@ -444,10 +448,10 @@ def main(C, OUT, IN, dt, grid, phys):
                 OUT["subT"],
                 OUT["Tsurf"],
                 kk_sz_top,
-                kk_sz_mid,
+                kk_sz_interior,
                 dz1,
                 dz2,
-                denom_l1,
+                denom_layer1,
                 denom_interior,
                 denom_bottom,
                 dt_stab,
@@ -478,12 +482,12 @@ def main(C, OUT, IN, dt, grid, phys):
                     break
                 # Calculate vertical heat fluxes
                 kdTdz[idx, 1] = kk_sz_top[idx] * (T_old[idx, 1] - OUT["Tsurf"][idx]) / dz1[idx]
-                kdTdz[idx, 2:] = kk_sz_mid[idx] * (T_old[idx, 2:] - T_old[idx, 1:-1]) / dz2[idx]
+                kdTdz[idx, 2:] = kk_sz_interior[idx] * (T_old[idx, 2:] - T_old[idx, 1:-1]) / dz2[idx]
 
                 # Update layer-wise temperatures
                 C_day_dt = C["dayseconds"] * dt_temp[idx]
 
-                T_new[idx, 1] = T_old[idx, 1] + C_day_dt * (kdTdz[idx, 2] - kdTdz[idx, 1]) / denom_l1[idx]
+                T_new[idx, 1] = T_old[idx, 1] + C_day_dt * (kdTdz[idx, 2] - kdTdz[idx, 1]) / denom_layer1[idx]
 
                 T_new[idx, 2:-1] = (
                     T_old[idx, 2:-1]
--- a/src/ebfm/core/LOOP_SNOW.py
+++ b/src/ebfm/core/LOOP_SNOW.py
@@ -17,6 +17,8 @@ except NameError:
 
 from ebfm.core import logging
 
+_SUCCESS = True
+
 logger = logging.getLogger(__name__)
 
 
@@ -128,7 +130,7 @@ def main(C, OUT, IN, dt, grid, phys):
             # Update runoff for shifted layers
             OUT["runoff_irr_deep"][is_shift] += subW_old[is_shift, nl - 1]
 
-        return True
+        return _SUCCESS
 
     def melt_sublimation():
         """
@@ -204,7 +206,7 @@ def main(C, OUT, IN, dt, grid, phys):
                     OUT["subZ"][idx, nl - 1] = grid["max_subZ"]
                 OUT["subD"][idx, nl - 1] = subD_old[idx, nl - 1]
 
-        return True
+        return _SUCCESS
 
     @profile
     def compaction():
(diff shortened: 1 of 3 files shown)


src/ebfm/core/LOOP_SNOW.py
matthias-ibach, line 375: Previous comment:
"I don't know what this does. I assume you do not need a comment here if the reader knows that the code does. Could you elaborate briefly here in the review?"

Answer:
"The `np.errstate()` context is needed because `gamma_drift` can be zero where `SI ≤ 0`. This would produce a divisionby-zero warning when computing `tau_i`. The zero gamma_drift points are always filtered out by `cond_drift_total` (which requires `SI>0) when the results is actually used, so the divide-by-zero values are harmless. The `errstate` simply suppresses the NumPy warnings for those elements.

matthias-ibach, line 429: Previous comment:
"Is is correct to start with the index 1: here? Why exclude OUT["subZ"][:, 0] etc.?"

Answer:
Yes, the layer 0 is a surface ghost layer and temperature is overwritten from Tsurf at end of this function, so it is excluded from the CFL condition.
A short comment has been added to the code to make this clear.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -427,11 +428,14 @@ def heat_conduction():
 
         # subZ and c_eff do not change
         # Precompute kk*subZ products once
+        # kk_sz_top: conductivity-thickness product for the top interface
+        # kk_sz_interior: same for all interior interfaces
         kk_sz_top = kk[:, 0] * OUT["subZ"][:, 0] + 0.5 * kk[:, 1] * OUT["subZ"][:, 1]
-        kk_sz_mid = kk[:, 1:-1] * OUT["subZ"][:, 1:-1] + kk[:, 2:] * OUT["subZ"][:, 2:]
+        kk_sz_interior = kk[:, 1:-1] * OUT["subZ"][:, 1:-1] + kk[:, 2:] * OUT["subZ"][:, 2:]
```
matthias-ibach, line 436: Previous comment:
"...The kk_sz_top / kk_sz_mid naming makes the situation quite clear. (even though _mid is an odd name because it's not only the middle layers but all layers except the top layer?)..."

Answer:
`kk_sz_mid` was renamed to `kk_sz_interior` at all locations. Also, the comment on the definition was updated to clarify it covers all interfaces except the top.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -427,11 +428,14 @@ def heat_conduction():
...
 
         # Precompute full temperature-update denominators once
-        denom_l1 = c_eff[:, 1] * (0.5 * OUT["subZ"][:, 0] + 0.5 * OUT["subZ"][:, 1] + 0.25 * OUT["subZ"][:, 2])
+        # denom_layer1: first active layer (layer 0: surface ghost layer overwritten from Tsurf)
+        denom_layer1 = c_eff[:, 1] * (0.5 * OUT["subZ"][:, 0] + 0.5 * OUT["subZ"][:, 1] + 0.25 * OUT["subZ"][:, 2])
```
matthias-ibach, line 440: Previous comment:
"What is _l1? _layer1? Is it _layer1 or actually _layer0? To not get too deep into indexing confusion: Maybe again use something like top (if it is the top)."

Answer:
`denom_l1` was renamed to `denom_layer1` with a short comment as well.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: Previous comment:
"What does True mean here (and also at other places)? Maybe use a constant SUCCESS = True here to improve readability."

Answer:
One problem with the `return True` that came to my mind here in general: since none of the return values are actually checked at the call sites (e.g., snowfall_and_deposition() is called without capturing the return), the `return True` is effectively dead. 
You could either delete the `return True` statements where they are actually meaningless (not the case in `layer_merging_and_splitting()` ), but I decided to keep them for the sake of having some visual reference for the end of the function. 

Or what do you prefer: Doing the  `_SUCCESS`  constant and maybe thinking of a way of checking the return values, but maybe rather with raising exceptions and not return codes,  just dropping the returns or leaving it as is?

src/ebfm/main.py
```diff
@@ -347,10 +349,10 @@ def _main_impl():
         help="If provided, log output will be written to the specified file (one file per MPI rank).",
     )
 
-    diag_group = parser.add_argument_group("diagnostics, reference snapshots, random seed")
+    diagnostics_group = parser.add_argument_group("diagnostics, reference snapshots, random seed")
```
matthias-ibach, line 352: Previous comment:
"diag could also be misunderstood as "diagonal" (which does not make any sense but requires some thinking effort on the reader's side)"

Answer:
I renamed `diag_group` to `diagnostics_group` at the corresponding locations.

src/ebfm/main.py
```diff
@@ -491,7 +493,7 @@ def _main_impl():
 
     if args.random_seed is not None:
         np.random.seed(args.random_seed)
-        logger.info(f"[DIAG] NumPy random seed fixed to {args.random_seed} for reproducibility.")
+        logger.info(f"NumPy random seed fixed to {args.random_seed} for reproducibility.")
```
matthias-ibach, line 496: Previous comment:
"If you think you need a prefix here this indicates that you could also move the functionality into a different module. Then the logger will automatically prefix the message with the name of the respective file.

You could, for example, do

```
import ebfm.diagnostics

if args.random_seed is not None:
    diagnostics.set_random_seed(args.random_seed)
``` 

and implement the actual functionality in ebfm.diagnostics"

Answer:
I just dropped the tag for the moment and did not add a new module. This issue is better addressed via the dedicated `diag_logger` + `_format_stats` helper (see other comments).

src/ebfm/main.py
```diff
@@ -559,8 +561,8 @@ def _main_impl():
         # Calculate surface mass balance
         OUT = LOOP_mass_balance.main(OUT, IN, C)
 
-        if args.diagnostics:
-            print_diagnostics(logger, grid, OUT, t)
+        if args.print_diagnostics:
+            print_diagnostics(grid, OUT, t)
```
matthias-ibach, line 565: Previous comment:
"?

if the argument only enables printing I would make this more explicit. Currently, one could misunderstand this that everything w.r.t diagnostics is disabled.
"

Answer:
I renamed `--diagnostics` to `--print-diagnostics` and updated the calls and references accordingly. This should make it clearer.

src/ebfm/main.py
```diff
@@ -157,24 +164,19 @@ def dump_reference(logger, OUT, filepath: str):
...
-        logger.info(f"[DIAG t={t + 1}] smb:            min={smb.min():.4e}  max={smb.max():.4e}  mean={smb.mean():.4e}")
+        diag_logger.info(f"[t={t + 1}] smb:            {_format_stats(smb)}")
     if smb_cum is not None:
-        logger.info(
-            f"[DIAG t={t + 1}] smb_cumulative: "
-            f"min={smb_cum.min():.4e}  "
-            f"max={smb_cum.max():.4e}  "
-            f"mean={smb_cum.mean():.4e}"
-        )
+        diag_logger.info(f"[t={t + 1}] smb_cumulative: {_format_stats(smb_cum)}")
```
matthias-ibach: Previous comment:
"You are duplicating some code here. Maybe use a small helper function? print_smb_stats(smb)/print_smb_stats(smb_cum) that returns the string min=... max=... mean=... with the formatting you want?
"

Answer: I added a  `_format_stats(arr: np.ndarray) -> str` helper returning returning the required information.

src/ebfm/main.py
```diff
@@ -157,24 +164,19 @@ def dump_reference(logger, OUT, filepath: str):
...
+def print_diagnostics(grid, OUT, t):
     """Log key diagnostic values each timestep for performance and correctness analysis."""
 
     gpsum = grid.get("gpsum", "N/A")
     has_shading = grid.get("has_shading", False)
     smb = OUT.get("smb")
     smb_cum = OUT.get("smb_cumulative")
 
-    logger.info(f"[DIAG t={t + 1}] gpsum={gpsum}, shading={'on' if has_shading else 'off'}")
+    diag_logger.info(f"[t={t + 1}] gpsum={gpsum}, shading={'on' if has_shading else 'off'}")
```
matthias-ibach: Previous comment:
"I think now I see the reason for introducing the `[DIAG]` label. I believe you can introduce an independent logger for this purpose. This allows you also to apply formatting rules (like adding this label) to just that logger. This would allow you to call `diagnostics_logger.debug(...)` for diagnostics stuff and `logger.debug(...)` in all other situations.

Maybe there is also a nice way to avoid the need to always write DIAG t={t+1} if you introduce a dedicated logger. But I'm not sure..."

Answer:
The following was done:

1. Added a module-level `diagnostics_logger: Logger` type annotation alongside `logger`
2. Assigned a diagnostics_logger = getLogger("ebfm.diagnostics")` both times `logger` is (re-)assigned in `_main_impl`
3. Removed the `logger` parameter from `print_diagnostics`. It now uses `diagnostics_logger` from module scope
4. Replaced the `[DIAG t={t+1}] with [t={t+1}]`. The logger name `ebfm.diagnostics` already identifies these lines in log output

src/ebfm/core/LOOP_SNOW.py
BenjaminRodenberg: Generally, the error treatment depends on how fatal an error actually is. You can raise an exception that the user has to catch to prevent a failure or simply log a warning if the error is acceptable. Working with error codes is in my opinion generally a bit odd in Python but acceptable if that's the current state of the interface and you do not want to revise it in this PR. Replacing the bool `True` with an enum `SUCCESS` would improve readability without changing the interface. But in the end it depends on how much you want to "escalate" this PR ;)

BenjaminRodenberg (approved):
Looks good to me 👍 

Feel free to skip the improvements on error treatment discussed in the last open comment. Still: Reworking error treatment would not harm but I'm totally fine if this is a topic for a different PR and we merge this one.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: You are correct that error codes like this are not really Python-style. 

I still went ahead and have now replaced the `return True` with a named constant `_SUCCESS = True` to improve readability without changing the interface, as you suggested.

A more thorough revision with raising exceptions would be a more significant change to the interface and is probably better addressed in a dedicated PR.

PR #128, opened 2026-06-18
Realistic test case

wardvp:
# Greenland test case with CARRA2 forcing 1990-1999: 

### UPDATE - 29/06

- [x] Remove all .nc forcing and grid files from the repository
- [x] Upload two packages to Google Drive for coarse- and high-resolution tests:
  - [x] [`Greenland_one_year_lowres.zip`](https://drive.google.com/file/d/1bZUKxQ7QgQjwwPA8jTHqlJ-mXhwsU_OP/view?usp=sharing) [0.5 GB]
  - [x] [`Greenland_one_year_highres.zip`](https://drive.google.com/file/d/1fTFMt0QzGTrQFfzOmBC4xk1OonFsH9ZQ/view?usp=sharing) [6.6 GB]
- [ ] Create third test case of one month on coarse resolution mesh:
  - [x] Create forcing files (`1990_1_month.nc` and `1990_2_month.nc`) & add to `/examples/`.
  - [x] Copy `Greenland_grid_coarse.nc` to `Greenland_grid_1_month.nc`
- [x] Update `README.md` with running instructions for the three cases.
- [x] Update `CHANGELOG.md`.

Run May 1990 test case like this:
```python
ebfm --greenland-mesh examples/Greenland_grid_one_month.nc --with-numba --numba-threads 4 --start-time "1-May-1990 00:00" --end-time "1-Jun-1990 00:00"
``` 

-------------------------------------------------------------------------------------

### Prepare grid and forcing files:
- [x] Download CARRA2 data [_IN PROGRESS; 1990 & 1991 DONE_].
- [x] Pre-process CARRA2 data to create vectorized forcing files (one per half year). This reduces file size and facilitates much faster reading in `LOOP_climate_forcing`. 
- [x] Forcing files for a high-resolution test are on an external server: `1990_1.nc` (download [here](https://drive.google.com/file/d/1ek8mQJZlxe8aYzaS5GiZuHXb_NtI4QlS/view?usp=sharing); 4 GB) for Jan-Jun 1990, `1990_2.nc` (download [here](https://drive.google.com/file/d/1Frl2adPN1MdyzqRZlh8W6v_TD8azFec2/view?usp=sharing); 4 GB) for Jul-Dec 1990 etc. Users need to download the forcing files to the `/examples/` folder.
- [x] Prepare grid file for high-resolution test (`Greenland_grid.nc`) containing a mask with glacier fraction (`mask`) and grid coordinates (`latitude `and `longitude`) and elevations (`orography`), and add to `/examples/`.
- [x] Forcing files for low-resolution test (`1990_1_coarse.nc` and `1990_2_coarse.nc`) added to `/examples/`.
- [x] Prepare grid file for low-resolution test (`Greenland_grid_coarse.nc`) and add to `/examples/`.

### In `main`, `config` and `grid`:
- [x] Create command line argument `--greenland-mesh` Greenland_grid.nc and modify `main.py` and `config.py` accordingly.
- [x] In `config.py` change the timezone to `dt_UTC` to -3 for Greenland.
- [x] In `grid.py` create a new `GridInputType` named `GREENLAND`.

### In `LOOP_write_to_file`:
- [x] Update `LOOP_write_to_file` to select variables (exclude subsurface variables for now to reduce filesize) and set `io['freqout']` to 8 to save daily (averaged/total) fields. 

### In `INIT`:
- [x] Create function `INIT.read_Greenland_grid` which reads coordinates, elevations and glacier fraction from `Greenland_grid.nc`. All cells with a glacier fraction >0 will be modelled (mask set to 1). 
- [x] Implement changes in `INIT.init_grid` to individually process `grid["x_2D"]`, `grid["y_2D"]`, `grid["z_2D"]`, `grid["lon_2D"]`, `grid["lat_2D"]` for the MATLAB and Greenland cases.

### In `LOOP_climate_forcing`:
- [x] Create a new function `read_Greenland_data`, which reads all required meteorological fields from the vectorized forcing files for the (nearest) time-step.
- [x] In `read_Greenland_data` enable reading multiple timesteps at once (hard-coded to 16 for now) to reduce runtime associated with reading forcing data.
- [x] Updates to `main` to: 1) make sure that random weather is read for the MATLAB case and that read_Greenland_data is called for the  Greenland case; 2) calculate RH and VP for the Greenland case (as in the ICON case).

### Documentation:
- [ ] Inline comments in new code sections.
- [x] Update README with instructions on how to run the test case and visualize the output.
- [x] Add an entry under *develop* in the [CHANGELOG.md](https://github.com/EBFMorg/EBFM/blob/main/CHANGELOG.md) (may be skipped for minor changes not observable for the user) 

### Tests:
- [x] Run 1-Jan-1990 to 3-Jan-1990 with 2 numba kernels --- **SUCCESS**
- [x] Run 1-Jan-1990 to 3-Jan-1990 with 4 numba kernels --- **SUCCESS: Is marginally faster than with 2 kernels**
- [x] Run 1-Jan-1990 to 31-Dec-1990 with 4 numba kernels at 10-km resolution. --- **SUCCESS: takes ~7 minutes with 4  numba kernels**
- [ ] Run 1-Jan-1990 to 31-Dec-1990 with 4 numba kernels at 2.5-km resolution.  

### Example run command:

High-resolution test:
```python
ebfm --greenland-mesh examples/Greenland_grid.nc --with-numba --numba-threads 4 --start-time "1-Jan-1990 00:00" --end-time "1-Jan-1991 00:00"
```

Coarse-resolution test (no external forcing files needed):
```python
ebfm --greenland-mesh examples/Greenland_grid_coarse.nc --with-numba --numba-threads 4 --start-time "1-Jan-1990 00:00" --end-time "1-Jan-1991 00:00"
``` 
 
# Development todos

- [x] merge #138 
- [x] merge #142


src/ebfm/core/config.py
```diff
@@ -186,12 +186,16 @@ def __init__(self, args: Namespace):
...
             raise Exception("Missing grid.")
 
-        if args.elmer_mesh and args.matlab_mesh:
-            logger.error("Please provide either --elmer-mesh or --matlab-mesh, not both.")
+        if args.elmer_mesh and (args.matlab_mesh or args.greenland_mesh):
+            logger.error("Please provide either --elmer-mesh or --matlab-mesh or --greenland-mesh, not several.")
+            raise Exception("Invalid grid configuration.")
+
+        if args.matlab_mesh and args.greenland_mesh:
+            logger.error("Please provide either --elmer-mesh or --matlab-mesh or --greenland-mesh, not several.")
```
BenjaminRodenberg:
```suggestion
        if sum((args.elmer_mesh, args.matlab_mesh, args.greenland_mesh)) > 1:
            logger.error("Please provide either --elmer-mesh or --matlab-mesh or --greenland-mesh, not several.")
```

Easier to read and extend.

You could even introduce `selected_grids = (args.elmer_mesh, args.matlab_mesh, args.greenland_mesh)` above and use `any(selected_grids)` in the check that raises `Exception("Missing grid.")`.

src/ebfm/main.py
```diff
@@ -226,13 +224,13 @@ def _main_impl():
     input_group.add_argument(
         "--elmer-mesh",
         type=Path,
-        help="Path to the Elmer mesh file. Either --elmer-mesh or --matlab-mesh is required.",
+        help="Path to the Elmer mesh file. Either --elmer-mesh or --matlab-mesh or --greenland-mesh is required.",
```
BenjaminRodenberg:
```suggestion
        help=f"Path to the Elmer mesh file. {allowed_mesh_options}",
```

This part is becoming a bit repetitive. I would suggest to define a shared string `allowed_mesh_options="Either --elmer-mesh or --matlab-mesh or --greenland-mesh is required."` at a central place and reuse it here.

src/ebfm/main.py
```diff
@@ -226,13 +224,13 @@ def _main_impl():
...
     )
 
     input_group.add_argument(
         "--matlab-mesh",
         type=Path,
-        help="Path to the MATLAB mesh file. Either --elmer-mesh or --matlab-mesh is required.",
+        help="Path to the MATLAB mesh file. Either --elmer-mesh or --matlab-mesh or --greenland-mesh is required.",
```
BenjaminRodenberg:
```suggestion
        help=f"Path to the MATLAB mesh file. {allowed_mesh_options}",
```
See above.

src/ebfm/main.py
```diff
@@ -242,6 +240,13 @@ def _main_impl():
         " If --netcdf-mesh is provided elevations will be read from the given NetCDF mesh file.",
     )
 
+    input_group.add_argument(
+        "--greenland-mesh",
+        type=Path,
+        help="Path to the mesh file and climate data for a test on Greenland. "
+        "Either --elmer-mesh or --matlab-mesh or --greenland-mesh is required.",
```
BenjaminRodenberg:
```suggestion
        f"{allowed_mesh_options}",
```
See above.

.gitignore
```diff
@@ -10,3 +10,13 @@ Output/*.nc
...
 .tox
+examples/1990_1_forcing_vectorized.nc
+examples/1990_2_forcing_vectorized.nc
+examples/preprocess_greenland_forcing_files.py
+examples/add_mask_field_to_greenland_grid.py
+.idea/EBFM_with_YAC.iml
+.idea/inspectionProfiles/profiles_settings.xml
+.idea/misc.xml
+.idea/vcs.xml
+.idea/workspace.xml
```
BenjaminRodenberg:
```suggestion
```

You can set up git locally to ignore these files. Or just ignore the whole folder?

.gitignore
```diff
@@ -10,3 +10,13 @@ Output/*.nc
 Reboot/*.nc
 .coverage
 .tox
+examples/1990_1_forcing_vectorized.nc
+examples/1990_2_forcing_vectorized.nc
+examples/preprocess_greenland_forcing_files.py
+examples/add_mask_field_to_greenland_grid.py
```
BenjaminRodenberg: Does the test case work without these files?

.gitignore
```diff
@@ -10,3 +10,13 @@ Output/*.nc
...
+.idea/EBFM_with_YAC.iml
+.idea/inspectionProfiles/profiles_settings.xml
+.idea/misc.xml
+.idea/vcs.xml
+.idea/workspace.xml
+examples/download_CARRA2.py
```
BenjaminRodenberg: If this is a script that could be useful for others, we could add it in the [tools](https://github.com/EBFMorg/EBFM/tree/main/tools) folder. If it is unrelated, maybe consider removing it or adding it in a different PR.

src/ebfm/core/LOOP_write_to_file.py
```diff
@@ -40,11 +40,11 @@ def main(OUTFILE, io, OUT, grid, t, time):
...
-            ["subD", "kg m^-3", "sample", "Density"],
-            ["subT", "K", "sample", "Temperature"],
-            ["subS", "mm w.e.", "sample", "Slush water content"],
-            ["subW", "mm w.e.", "sample", "Irreducible water"],
-            ["subZ", "m", "sample", "Layer thickness"],
+            # ["subD", "kg m^-3", "sample", "Density"],
+            # ["subT", "K", "sample", "Temperature"],
+            # ["subS", "mm w.e.", "sample", "Slush water content"],
+            # ["subW", "mm w.e.", "sample", "Irreducible water"],
+            # ["subZ", "m", "sample", "Layer thickness"],
```
BenjaminRodenberg: Why did you have to comment this out? Looks to me like we have to make this configurable depending on the mesh type?

BenjaminRodenberg (changes requested):
Some minor comments are left. But generally looks good.

I would suggest to also add this case in the CI pipeline to make sure it also works in the future.

Another thing: I think it would be helpful to add a (very short) section/paragraph in the `README.md` to tell users this case exists and how they run it (would also help me trying out the case to test it before merging).

src/ebfm/core/config.py
```diff
@@ -304,7 +314,7 @@ def __init__(self, args: Namespace):
                 f"day ({SECONDS_PER_DAY} seconds). This may lead to unexpected behavior."
             )
 
-        self.dT_UTC = 1  # Time difference relative to UTC in hours (hard-coded for now)
+        self.dT_UTC = -3  # Time difference relative to UTC in hours (hard-coded for now)
```
BenjaminRodenberg: Why is this needed? Something we need to make configurable or can we infer this from (for example) the forcing files?

src/ebfm/core/LOOP_climate_forcing.py
```diff
@@ -43,7 +47,10 @@ def main(C, grid, IN, t, time, OUT, cpl: Coupler) -> tuple[dict, dict]:
     # SPECIFY/READ METEO FORCING
     ###########################################################
     if not cpl.has_coupling_to("icon_atmo"):
-        IN = set_random_weather_data(IN, C, time, grid)
+        if config.grid_type is GridInputType.GREENLAND:
+            IN = read_Greenland_data(IN, C, time, grid, config)
+        else:
+            IN = set_random_weather_data(IN, C, time, grid)
```
BenjaminRodenberg: Not sure how standardized these forcing files are, but why not implement this differently:

```sh
if args.forcing_file:  # forcing file provided by user
    IN = read_Greenland_data(...)  # probably need to use a more generic name here
else:
    IN = set_random_weather_data(...)
```

I imagine it would be useful to apply the forcing files not only to the new greenland mesh but also to the elmer or MATLAB mesh.

src/ebfm/core/LOOP_climate_forcing.py
```diff
@@ -100,6 +109,135 @@ def main(C, grid, IN, t, time, OUT, cpl: Coupler) -> tuple[dict, dict]:
...
+        dict: Updated IN dictionary with meteorological data.
+    """
+    forcing_dir = config.mesh_file.parent
+
+    model_time_utc = time["TCUR"] - timedelta(hours=time["dT_UTC"])
+    model_time_seconds = model_time_utc.replace(tzinfo=timezone.utc).timestamp()
+
+    file_group = 1 if model_time_utc.month <= 6 else 2
+    forcing_prefix = f"{model_time_utc.year}_{file_group}"
+    forcing_file = forcing_dir / f"{forcing_prefix}_forcing_vectorized.nc"
```
BenjaminRodenberg: Rather than hard-coding the path to the forcing file I would let the user provide the path.

Why do you have to differentiate between `1990_1_forcing_vectorized.nc` and `1990_2_forcing_vectorized.nc`? I'm guessing you had to split the file because a single file would have been too large?

Instead of asking the user to provide multiple files, maybe you could put them in a folder and let the user provide the path to the folder?

```
ebfm ... --forcing-files examples/greenland_forcing
```

The folder could be organized as follows:

```
greenland_forcing/
├── 1990-01.nc
├── 1990-02.nc
...
└── 1990-12.nc
```

BenjaminRodenberg (commented):
A few comments on the treatment of forcing files.

examples/Greenland_grid.nc.license
```diff
@@ -0,0 +1,3 @@
+# SPDX-FileCopyrightText: TODO
```
matthias-ibach: What should be there at the "TODO"?
`# SPDX-FileCopyrightText: 2026 The EBFM Authors` ? 
Or the actual authors of the grid?

src/ebfm/core/config.py
```diff
@@ -240,8 +248,10 @@ def __init__(self, args: Namespace):
             )
             raise Exception("Invalid grid configuration.")
 
-        # Shading is only supported for MATLAB meshes; see https://github.com/EBFMorg/EBFM/issues/11
-        grid_type_supports_shading_supported = self.grid_type is GridInputType.MATLAB
+        # Shading is only supported for MATLAB & Greenland meshes; see https://github.com/EBFMorg/EBFM/issues/11
```
matthias-ibach: Is this comment still relevant? Since #11 has been closed?

src/ebfm/core/INIT.py
```diff
@@ -410,7 +422,8 @@ def init_grid(grid: GridDict, io, config: GridConfig):
             max_angle = np.full(grid["gpsum"], -np.inf, dtype=np.float64)
             count = 1
             active = np.ones(grid["gpsum"], dtype=bool)
-            while active.any():
+
+            while active.any() and count * grid["dx"] < 5e4:
```
matthias-ibach: Why `< 5e4` ?

README.md
```diff
@@ -78,9 +78,17 @@ After installation, a basic, uncoupled simulation can be run with the following
...
 ```
 
+A more advanced case can be run via
+
+```sh
+ebfm --greenland-mesh examples/Greenland_grid.nc --start-time "01-Jan-1990 00:00" --end-time "31-Dec-1990 00:00"
+```
+
+Note: Please download [`1990_1_forcing_vectorized.nc`](https://drive.google.com/file/d/13pIjbIljL4LUPuOtBm0ArX23NCwgLhOZ/view?usp=sharing) and [`1990_2_forcing_vectorized.nc`](https://drive.google.com/file/d/13tVXkXDA1Nhf_X-IK9qE_hEwIv1wme9x/view?usp=sharing) and put these files into the `examples` folder before running the case.
+
```
matthias-ibach, line 108: Should we add a note about where the output is saved and how to view it, e.g. mentioning ncview?
Something like: "The output of the run is saved to `/Output/model_output.nc` and can be viewed with e.g. `ncview`." (or maybe not reference `ncview` as not to confuse the user or having to add new instructions on how to use `ncview`).

src/ebfm/core/config.py
BenjaminRodenberg: I see the point. I think the current situation is not so clear and it's also hard to judge the current state without tests. I would suggest to keep the shading support as it is here and revise it in a future PR.

I opened #129 and also revised the implementation of shading + grid config here to reduce the amount of code duplication and risk of forgetting updates.

src/ebfm/main.py
BenjaminRodenberg: Took care of this part in https://github.com/EBFMorg/EBFM/pull/128/commits/69d2a2ebafff25a990268273476823e584988b9c.

src/ebfm/main.py
BenjaminRodenberg: Took care of this part in https://github.com/EBFMorg/EBFM/pull/128/commits/69d2a2ebafff25a990268273476823e584988b9c.

src/ebfm/main.py
BenjaminRodenberg: Took care of this part in https://github.com/EBFMorg/EBFM/pull/128/commits/69d2a2ebafff25a990268273476823e584988b9c.

README.md
BenjaminRodenberg: I added a note on `ncview` and decided to remove it again. I thought that users of this code will know how to visualize and interpret with a netCDF file. I would say let's keep it as it is now. If at some point we realize that we actually need to look at the output more frequently and this causes confusion, we can still clarify the steps in the docs.

src/ebfm/core/INIT.py
BenjaminRodenberg: True. The `5e4` looks to me like a "magic number". Introducing a named variable and (if needed) adding a comment should be the right way to go here.

examples/Greenland_grid.nc.license
BenjaminRodenberg: If this file has been derived from some other file, you could also point to the original using a comment. Compare to https://github.com/EBFMorg/EBFM/blob/realistic-test-case/examples/BedMachineGreenland-v5_lo.nc.license

src/ebfm/core/config.py
BenjaminRodenberg: Cleaned this up in https://github.com/EBFMorg/EBFM/pull/128/commits/6ac76458973ff56300cb209594e81fd91bd19ea5.

.gitignore
wardvp: Thanks, I will try to solve this locally.

.gitignore
wardvp: The test case needs the two .nc files, but the two .py files are not needed in the repository. I only use them locally to produce the final Greenland_grid.nc and the two vectorized forcing .nc files.

.gitignore
wardvp: This script does not need to be in the repository. I did not even use it myself, as I downloaded the forcing data via the web portal.

src/ebfm/core/LOOP_write_to_file.py
wardvp: That was just a temporary fix to reduce the file size of the output file, and for faster file writing during tests.

src/ebfm/core/INIT.py
wardvp: Yes, this magic number needs some description (and a variable name). It is the search distance to check for shading. I introduced it because on a large grid like Greenland it does not make sense to walk all the way to the edges of the grid to check for shading. A search distance of 50 km (5e4 m) was set, which makes the calculations much faster and almost identical to the original method.

src/ebfm/core/config.py
wardvp: Eventually, I think this can be removed. Right now, it is used because the model time is actual (local) time which for Greenland is at UTC-3. Then in the model code it is made sure that the meteorological fields are read for the correct local time (the forcing files have times in UTC) and that solar insolation also corresponds to the same time. But in the end it may make more sense to have just one time in the model, which is the time in UTC. This will require a few changes in LOOP_climate_forcing and LOOP_EBM_insolation.

src/ebfm/core/LOOP_climate_forcing.py
wardvp: Good idea, I will change this. For now, I will just change the names to 1990_1 and 1990_2. Later they can also be put in a separate folder, but not sure yet as it adds yet another command line argument that could also be avoided.

examples/Greenland_grid.nc.license
wardvp: The grid is, like the forcing files, downloaded from the Copernicus CARRA2 server. The website indicates a CC-BY licence. Reference: https://doi.org/10.24381/f5effe24

src/ebfm/core/config.py
BenjaminRodenberg: Is this something you think that is doable in this PR or better postpone it and open an issue? Especially for global coupled simulations I guess timezones can be quite a headache (similar to calendars #127).

.gitignore
BenjaminRodenberg: I added some files to my local ignore here:

```
EBFM$ cat .git/info/exclude 
# git ls-files --others --exclude-from=.git/info/exclude
# Lines that start with '#' are comments.
# For a project mostly in C, the following would be a good set of
# exclude patterns (uncomment them if you want to use them):
# *.[oa]
# *~

examples/elmer_w_DEM
examples/elmer_wo_DEM
examples/DEM
examples/BedMachine*

# local excludes

examples/1990_1_forcing_vectorized.nc
examples/1990_2_forcing_vectorized.nc
.venv*
*.orig
```

.gitignore
BenjaminRodenberg: I think it would make sense to keep the Python files around, for example in tools. It helps if at some point we are wondering how these files were created. I also added some similar scripts for reproducibility reasons.

You could also add a short section in [`tools/README.md`](https://github.com/EBFMorg/EBFM/tree/main/tools#nc_reduce_sizepy---shrink-a-netcdf-file) that tells the user/developer how to create these files from the raw data.

.gitignore
BenjaminRodenberg: Personally, I prefer a script if it is possible to execute the workflow with a script. So if the script works: Let's keep it for reproducibility. I can also review this part of the workflow.

src/ebfm/core/LOOP_write_to_file.py
BenjaminRodenberg: Ok. So if you only needed this during development I assume we can just revert this?

BenjaminRodenberg: Took care of this in 9667342e763f5776597f883b835dd9eb1aefc3c5.

examples/Greenland_grid.nc.license
BenjaminRodenberg: I added license information in 936038cfac8badec5fb877505ae1cdca5cb78a73.

I followed the approach from #123 assuming both files are modified version of the ones from Copernicus. So: Copyright goes to "EBFM Authors" + a note "Derived from: ..." where we point to the original Copernicus dataset.

examples/1990_1_coarse.nc.license
BenjaminRodenberg, line 1: @wardvp can you point me to the original source for this file? If this is again a derived file I would add a copyright notice + derived from info like for the other files.

examples/1990_2_coarse.nc.license
BenjaminRodenberg, line 1: I assume same as for `1990_1_coarse.nc.license` applies here.

examples/Greenland_grid.nc
BenjaminRodenberg, line 1: Do we need both files `Greenland_grid.nc` and `Greenland_grid_coarse.nc` here?

I think it would be helpful to briefly describe the purpose of all files that you added here in `examples/README.md`.

BenjaminRodenberg (commented):
I did some updates to the license and a focused review on what's still missing there. @wardvp feel free to directly implement the required changes here or give feedback on the review comment and I can take care of the implementation (or ask if there are open questions).

src/ebfm/core/LOOP_climate_forcing.py
wardvp: Right now, the forcing files are specifically designed for the Greenland test case as they contain vectorized weather data for only the grid points that are modeled. I.e. the same forcing files can not be used to simulate on another grid. We could generalize this of course to read full CARRA2 forcing files, but they are simply too large for a test case.

src/ebfm/core/LOOP_climate_forcing.py
wardvp: Following up on this, I think it would be better to just hard-code the path. In the coarse-scale Greenland experiment the forcing files will come with the package when installing EBFM and will be in the examples folder. If the user wants to run the fine-scale experiment, the user will need to download the files from the external server and make sure to put them into the examples folder. That should not be too complicated and avoids having to specify another command line argument when running the model. The README should of course explain how to run the test cases.

examples/Greenland_grid.nc
wardvp: Yes, I need to add info to the README. When running with `--greenland-mesh examples/Greenland_grid.nc` it activates the high-resolution run (requires the externally downloaded 3-GB forcing files). Instead, when running with `--greenland-mesh examples/Greenland_grid_coarse.nc` the coarse-resolution Greenland test will start (reading instead the 250 MB forcing files).

CHANGELOG.md
```diff
@@ -30,7 +31,6 @@ SPDX-License-Identifier: CC-BY-4.0
   * Improvements in LOOP_SNOW.py (compaction, heat_conduction, percolation_refreezing_and_storage and layer_merging_and_splitting)
   * Added numba kernels for compaction, heat_conduction and percolation_refreezing_and_storage (in `LOOP_SNOW_kernels.py`) , addresses: https://github.com/EBFMorg/EBFM/issues/55
   * Introduced `compute_backend.py` to manage compute-backend dispatch to separate kernel code from logic. Explicit `if/else` dispatch with a single return per function. Prepares codebase for adding, e.g., GPU offload backends without structural changes.
-* Bug fixes in double depth method in INIT.py and LOOP_SNOW.py (by Ward)
```
wardvp: This one was listed twice...

examples/Greenland_grid.nc
BenjaminRodenberg: Is it also possible to use `Greenland_grid_coarse.nc` for both cases? A quick first test run with potentially physically incorrect results is ok. If a user wants the real physics it is fine to expect the user to download the high-resolution file in my opinion.

Having `Greenland_grid_coarse.nc` here is rather pointless when we also store `Greenland_grid.nc`; due to the file size I would prefer avoid storing `Greenland_grid.nc` here.

examples/Greenland_grid_coarse.nc.license
BenjaminRodenberg, line 1: Same as for `examples/Greenland_grid.nc.license`?

src/ebfm/core/LOOP_EBM_insolation.py
```diff
@@ -52,7 +52,7 @@ def main(grid, time2, OUT):
         9.87 * np.sin(2.0 * np.radians(B)) - 7.53 * np.cos(np.radians(B)) - 1.5 * np.sin(np.radians(B))
     )  # Correction for eccentricity
 
-    Tcor_lon = 4 * (grid["lon"] - 15 * time2["dT_UTC"])  # Correction for longitude within time-zone
+    Tcor_lon = 4 * grid["lon"]  # Correction for longitude within time-zone
```
BenjaminRodenberg, line 57: Is this correction not needed anymore in general? I think we should try to separate this part from the PR here and solve it independently. I'll quickly open a PR with the corresponding changes.

wardvp: Yes, it could go to a separate PR. The main change is that model time (in time["TCUR"]) is now always in UTC(+0), which makes that time-zone corrections like here are no longer needed.

src/ebfm/core/LOOP_climate_forcing.py
```diff
@@ -100,6 +109,149 @@ def main(C, grid, IN, t, time, OUT, cpl: Coupler) -> tuple[dict, dict]:
...
+        time (dict): Time-related variables.
+        grid (dict): Grid information.
+        config (GridConfig): Grid configuration.
+
+    Returns:
+        dict: Updated IN dictionary with meteorological data.
+    """
+    forcing_dir = config.mesh_file.parent
+
+    model_time_utc = time["TCUR"]
```
BenjaminRodenberg, line 127:
```suggestion
    model_time = time["TCUR"]
```
To me adding the suffix `_utc` looks like an implementation detail to me. `LOOP_climate_forcing.py` should not have to worry about this part.

src/ebfm/core/LOOP_climate_forcing.py
```diff
@@ -100,6 +109,149 @@ def main(C, grid, IN, t, time, OUT, cpl: Coupler) -> tuple[dict, dict]:
...
+    model_time_seconds = model_time_utc.replace(tzinfo=timezone.utc).timestamp()
```
BenjaminRodenberg, line 128: Maybe something we can put into a small helper function in the time config? We could also store it in the `time` dict; would be slightly redundant. But I see this rather as a responsibility of the time module.

examples/Greenland_grid_coarse.nc.license
wardvp: Indeed, it is just a coarse-resolution version of it.

BenjaminRodenberg (commented):
I created https://github.com/EBFMorg/EBFM/pull/138 to take care of the UTC topic independently. I think it is a bit confusing to solve this issue here.

examples/Greenland_grid.nc
wardvp: The two runs use the same physics, but use a different mesh, either a 10-km resolution one or a 2.5-km resolution one. That's why they also use different (vectorized) forcing files. So it is needed to have the separate input files. 

We could instead provide an external download link for Greenland_grid.nc though, like for the climate forcing files of the 2.5-km experiment. The file is 130 MB.

The reasons to have a coarse-resolution test is to be able to run the testcase without having to download external files (as everything is already in /examples/ after installing EBFM), and it is clearly faster than the high-resolution experiment.

src/ebfm/core/LOOP_EBM_insolation.py
BenjaminRodenberg: Ok. Let's more this information into #138.

CHANGELOG.md
BenjaminRodenberg: Cleaned this up https://github.com/EBFMorg/EBFM/commit/72e6a792b39071b00869b14301192a83fd793f93. Thanks for the hint!

examples/Greenland_grid.nc
BenjaminRodenberg: As far as I understand we can delete this file here when we put it on Google drive? I suggested a minor edit in the Readme.

examples/1990_2_coarse.nc
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:688f3d6e8800fc74c1a30e23331c5932748f1f68d16fc2c12b9e3a59cf7d8e3c
+size 259405806
```
BenjaminRodenberg: 250 MB is not particularly small... could we reduce the size by, for example, only running a single month? I assume the physics will be relatively uninteresting in this case but it's just for the sake of the example.

If a user wants a more realistic case it is necessary to download the files from Google Drive.

README.md
```diff
@@ -77,10 +77,29 @@ After installation, a basic, uncoupled simulation can be run with the following
...
+```sh
+ebfm --greenland-mesh examples/Greenland_grid_coarse.nc --start-time "01-Jan-1990 00:00" --end-time "1-Jan-1991 00:00"
+```
+
+or on a 2.5-km grid, after downloading forcing files [`1990_1.nc`; 4 GB](https://drive.google.com/file/d/1ek8mQJZlxe8aYzaS5GiZuHXb_NtI4QlS/view?usp=sharing) and [`1990_2.nc`; 4 GB](https://drive.google.com/file/d/1Frl2adPN1MdyzqRZlh8W6v_TD8azFec2/view?usp=sharing) and putting these files into the `examples` folder before running the case:
+
+```sh
+ebfm --greenland-mesh examples/Greenland_grid.nc --start-time "01-Jan-1990 00:00" --end-time "1-Jan-1991 00:00"
+```
```
BenjaminRodenberg: I'm currently facing the problem that I cannot download the files since we reached the monthly bandwith limit for git LFS. Could you also put the coarse resolution files on Google Drive? Mainly as a backup.

You could use the following description in the `README.md` (could not directly add a suggestion here because github got confused with me mixing code and suggestions). Feel free to edit. Especially, the "... generated from the CARRA dataset"-part might just be wrong?

---
A more advanced test case simulates the Greenland Ice Sheet during a full year (1990), and can be run on a 10-km coarse resolution grid [`Greenland_grid_coarse.nc`](TODO) using realistic forcing for one month generated from the CARRA dataset [`1990_1_coarse.nc`](TODO):

```sh
ebfm --greenland-mesh examples/Greenland_grid_coarse.nc --start-time "01-Jan-1990 00:00" --end-time "1-Jan-1991 00:00"
```

This run is relatively coarse resolution and short term and acts as a basic example that may be run out-of-the-box when downloading this repository.

For more realistic physics on a 2.5-km grid over one year, please download the fine-resolution grid file [`Greenland_grid.nc`; ~100 MB](TODO) and the corresponding forcing files [`1990_1.nc`; 4 GB](https://drive.google.com/file/d/1ek8mQJZlxe8aYzaS5GiZuHXb_NtI4QlS/view?usp=sharing) and [`1990_2.nc`; 4 GB](https://drive.google.com/file/d/1Frl2adPN1MdyzqRZlh8W6v_TD8azFec2/view?usp=sharing) and put these files into the `examples` folder. Then run the case with the following command:

```sh
ebfm --greenland-mesh examples/Greenland_grid.nc --start-time "01-Jan-1990 00:00" --end-time "1-Jan-1991 00:00"
```
---

README.md
```diff
@@ -72,16 +72,44 @@ If during the installation of EBFM it appears that `yac` is missing, please doub
...
+ebfm --netcdf-mesh examples/Greenland_CARRA2_fine/grid.nc --start-time 1990-01-01T00:00:00 --end-time 1991-01-01T00:00:00
+```
+
+for the 2.5-km resolution experiment.
+
+Please note that runs become significantly faster (~3 times) when running with performance optimization (see also **Performance and Profiling Runs** below), e.g.:
+
+```sh
+ebfm --netcdf-mesh examples/Greenland_CARRA2_fine/grid.nc --numba-threads 2 --start-time 1990-01-01T00:00:00 --end-time 1991-01-01T00:00:00
+```
```
BenjaminRodenberg: I updated the instructions to the new structure. Still needs testing and we can, of course, also change the naming to something else than `Greenland_CARRA2_fine`.

@wardvp Are the forcing files for the 2.5km and 10km resolution in principle the same? If yes, we could also go with the following structure:

```
examples/
├── ...
└── Greenland_CARRA2
    ├── ...
    ├── 1990-04-01      # requires download
    │   └── P1M.nc
    ├── 1990-05-01      # already present
    │   └── P1M.nc
    ├── 1990-06-01      # requires download
    │   └── P1M.nc
    ├── ...
    ├── ...
    ├── grid_coarse.nc  # already present
    └── grid_fine.nc    # requires download
```

BenjaminRodenberg:
Rebasing this branch onto `v0.7.0` was complicated. Here is a pointer to the latest commit before the rebase in case something broke and we need it https://github.com/EBFMorg/EBFM/pull/128/commits/b9b73beb4bc19fa7fa2134a3be61fdd725d8bba3

I did the rebase manually first and then refinement with Claude. I'm posting an AI generated summary. If something looks dubious @wardvp or @matthias-ibach, please let me know and we can look into it and fix it.

---

## Rebase onto v0.7.0 — resolution notes

Last sync was ~v0.5.0, so this rebase pulled in v0.6.0, v0.6.1 and v0.7.0.
Resolution policy: **released code in v0.7.0 wins; branch WIP adapts to it.**

Of main's 18 changelog entries, only one genuinely collided with branch work
(#155, the shading-LUT skip) — both sides edited the same block in `init_grid`.
Everything else landed cleanly; `--netcdf-dem-mesh` (#142) and this branch's new
`--netcdf-mesh` are distinct flags and coexist, and dT_UTC (#138) was removed on
both sides independently.

### Judgement calls made during resolution

1. **`read_NETCDF_grid` glacier mask** — restored `orography` / `mask` over the
   `orog` / `orog >= 10` variant that the rebase had reinstated. `git log -S`
   shows the latter is from the branch's *first* commit, superseded in the very
   next one, and absent from the branch tip. Note these are physically different
   glacier definitions: "file mask > 0 AND finite orography" vs "elevation ≥ 10 m".
   The rebased tree was also calling `orog`, which is undefined — `NameError` on
   every NetCDF grid read.

2. **Shading guard** — restored #155's `_precompute_shading_matlab` helper, gated
   on `config.grid_type is MATLAB and grid["has_shading"]`. The grid-type check is
   redundant today (`grid_types_supporting_shading == {MATLAB}`), kept deliberately
   to document the rotated-cell limitation for NetCDF grids.

3. **Kept the branch's `max_walk_distance = 5e4`** rather than reverting to main's
   unbounded walk to the domain edge. See open question below.

4. **Dropped a duplicated lat/lon block** that the rebase left running
   unconditionally. It re-derived lon/lat from x/y using `utmzone = 33`
   (EPSG:32633) while `read_NETCDF_grid` builds x/y with EPSG:32624 — a wrong-zone
   round trip that overwrote the NetCDF file's own lat/lon. Kept the per-grid-type
   assignment (computed for MATLAB, read from file for NetCDF).

5. **Scope expansion:** fixed a pre-existing branch bug,
   `is GridInputType.MATLAB or GridInputType.NETCDF`, which parses as
   `(x is MATLAB) or NETCDF` and is therefore always true, making the
   `else: raise ValueError` unreachable. Behaviour change: `CUSTOM`/`ELMERXIOS`
   without a separate DEM now raise the intended error instead of being fed to
   `read_NETCDF_grid`.

### Open questions

- **`max_walk_distance = 5e4` changes shading physics vs released v0.7.0** for
  MATLAB grids, with no changelog entry. Is the 50 km cap meant to ship? Also,
  `count * grid["dx"] < max_walk_distance` uses `dx` for both axes, so the
  effective cutoff is azimuth-dependent (a diagonal walk covers ~√2 more ground
  per step than the test assumes).
- Please sanity-check the restored mask semantics against the CARRA2 inputs (item 1).
- The `# develop` entry for `--netcdf-mesh` doesn't mention that shading is
  unavailable for NetCDF grids.
- `.gitattributes` was resolved by collapsing the per-file `.nc` LFS entries into a
  blanket `*.nc`, which now also captures anything under `examples/greenland_forcing/`.

BenjaminRodenberg:
I'm investigating the failure of the tests on the branch https://github.com/EBFMorg/EBFM/commit/e52b7201aa5602089a10a9a34fc9bad65ca2856f two possibilities:

* Problem with `dt_stab` was always there and only surfaces now due to the assertion
* Problem with `dt_stab` has been introduced by the rebase I did yesterday

BenjaminRodenberg:
@wardvp I analyzed the error a bit more:

* Added assertions via https://github.com/EBFMorg/EBFM/pull/128/changes/35c52ffe802d0eb12250b168ec86a2e53af847ca to make the failure mode more explicit. The commit message should already tell you quite well what's going on.
* According to AI there are some options on the table. I cannot really judge here and would leave this to you. Maybe the error if clear to you and you already have just the right fix in mind. If not: I can support you with what I learned during the debugging and the options the AI identified.
* I also added the new explicit assertions in the pre-rebase branch mentioned above. See: https://github.com/EBFMorg/EBFM/commit/bd0fc34c81b7a7ee2185091412808743ad914a84; here the CI runs into the same failure like here (see https://github.com/EBFMorg/EBFM/actions/runs/32476427530/job/96754004851). Just to be sure the rebase I did yesterday is not to blame.

wardvp:
> @wardvp I analyzed the error a bit more:
> 
>     * Added assertions via [35c52ff](https://github.com/EBFMorg/EBFM/commit/35c52ffe802d0eb12250b168ec86a2e53af847ca) to make the failure mode more explicit. The commit message should already tell you quite well what's going on.
> 
>     * According to AI there are some options on the table. I cannot really judge here and would leave this to you. Maybe the error if clear to you and you already have just the right fix in mind. If not: I can support you with what I learned during the debugging and the options the AI identified.
> 
>     * I also added the new explicit assertions in the pre-rebase branch mentioned above. See: [bd0fc34](https://github.com/EBFMorg/EBFM/commit/bd0fc34c81b7a7ee2185091412808743ad914a84); here the CI runs into the same failure like here (see https://github.com/EBFMorg/EBFM/actions/runs/32476427530/job/96754004851). Just to be sure the rebase I did yesterday is not to blame.

@BenjaminRodenberg Thanks for spotting this. Unfortunately, I am about to leave on fieldwork to northern Sweden. This issue will have to wait until I return, earliest on Friday next week. So the main issue is that dt_stab occassionally becomes negative? That can only happen when either the temperature (in Kelvin) or the density (in kg m-3) would locally be negative. Neither of those should of course happen. 

If you are willing to do some tests, you could try to see what happens if you change the 0.5 in the dt_stab formula to 0.25. If that solves the issue then the problem was that the conduction scheme locally does not converge (as it uses too large timesteps) which can lead to erroneous (possibly negative) temperatures. You could also check if indeed subsurface temperatures or densities locally become negative.

examples/README.md
```diff
@@ -31,3 +32,15 @@ python3 tools/nc_reduce_size.py \
...
 
 See `tools/README.md` for full documentation of `nc_reduce_size.py`.
+
+### How grid and forcing files in `Greenland_CARRA2` were produced
+
+Files in `Greenland_CARRA2` are organized in a specific way. Please refer to `Greenland_CARRA2/README.md` for details.
+
+#### Grid files `grid.nc`
+
+TODO
```
BenjaminRodenberg: Open TODO

examples/README.md
```diff
@@ -31,3 +32,15 @@ python3 tools/nc_reduce_size.py \
...
+
+#### Forcing files
+
+TODO
```
BenjaminRodenberg: Open TODO

BenjaminRodenberg (changes requested):
* Two documentation todos left
* The failure in the CI for CARRA2 still needs fixing

and then we can merge it.

PR #130, opened 2026-06-19
Introduce own module for CLI and tests

BenjaminRodenberg:
# Author's Todos

- [x] ~~Merge #128~~ removed dependency via rebase


src/ebfm/core/cli.py
```diff
@@ -0,0 +1,460 @@
+# SPDX-FileCopyrightText: 2026 EBFM Authors
+#
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# This file was generated with the help of AI tools.
```
matthias-ibach, line 5: Do you want to keep this comment? 
It is true and arguably good practice to be transparent about AI-assisted code.

In general, what about if code is heavily modified, I am wondering if the comment should then still be kept? At what point does it become "or own work"?

Happy to keep it either way, just had some thoughts about it.

src/ebfm/core/cli.py
```diff
@@ -0,0 +1,460 @@
...
+from ebfm.core import INIT
+from ebfm.core.comm import mpi_available
+from ebfm.core.config import Calendar, FieldValidationLevel, GridConfig, TimeConfig
+from ebfm.core.grid import GridInputType
+from ebfm.core.logger import log_levels_map
+
+# Module-level parser reference, set by parse_cli_args().
+# Allows post-parse validation functions to call parser.error() without
+# requiring callers outside this module to hold a reference to the parser.
+_parser: ArgumentParser | None = None
```
matthias-ibach, line 30: I sometimes wonder if the AI generated comments are a bit lengthy. I think here it might be ok, but sometimes that is the case.

src/ebfm/core/cli.py
```diff
@@ -0,0 +1,460 @@
...
+
+
+def _cli_error(msg: str) -> None:
+    """Call parser.error() on the module-level parser. Must be called after parse_cli_args()."""
+    assert _parser is not None, "_cli_error() called before parse_cli_args()"
+    _parser.error(msg)
+
+
+class CliDefaults(Enum):
+    START_TIME = datetime(1979, 1, 1, 0, 0)
```
matthias-ibach, line 40: Is it ok to define these defaults here?
I am just wondering, I might not be deep enough into the code structure here...

src/ebfm/core/cli.py
```diff
@@ -0,0 +1,460 @@
...
+
+    @param[in] parser the argument parser to add the coupling arguments to.
+    """
+
+    # Note: If you add arguments to this function, also update check_coupling_features.
+
+    coupling_group = parser.add_argument_group("coupling (requires YAC)")
+
+    coupling_group.add_argument(
+        "--couple-to-elmer-ice",
```
matthias-ibach, line 64: I am just wondering: in our testcase runscripts (`run_ebfm_*.sh`), we usually use `--couple-to-elmer` instead of  `--couple-to-elmer-ice` (also see Clemens repo). It seems to work just fine, because I think it might get picked up anyways. It would probably be good practice to update the runscripts accordingly to avoid any confusion in the future?

src/ebfm/core/cli.py
```diff
@@ -0,0 +1,460 @@
...
+    )
+
+    coupling_group.add_argument(
+        "--coupler-config",
+        type=Path,
+        help="Path to the coupling configuration file (YAC coupler_config.yaml).",
+    )
+
+    coupling_group.add_argument(
+        "--field-validation-level",
```
matthias-ibach, line 82: Should something like this be added to the README? So the users know about it? Or is it enough to have it here, where it can be seen via the `--help` CLI?

tox.ini
```diff
@@ -15,6 +15,14 @@ description = Run unit tests across Python versions
 commands =
     pytest tests/ -v --cov=src/ebfm --cov-report=term-missing
 
+[testenv:unit-mpi]
+description = Run unit tests with MPI available (includes MPI-guarded tests)
+extras =
+    dev
```
matthias-ibach: Maybe not necessary, if inherited from `[testenv] extras = dev` above?

matthias-ibach (commented):
Just some very general comments that do not necessarily need to be addressed, just something to discuss.

Other than that, this PR looks good to me. I have just skipped through the tests and have not looked at each one in detail or if every possible (feasible) combination is tested.

src/ebfm/core/cli.py
BenjaminRodenberg: > Do you want to keep this comment?

I added this comment intentionally. I will explain why below. But this is certainly a topic for discussion.

>  At what point does it become "or own work"?

I think this is a very controversial question ;) "This file was generated with the help of AI tools." does not make any statement about original work or copyright (from my non-lawyer perspective). According to the header this file is still copyrighted to "The EBFM Authors" and I think this also makes sense.

> In general, what about if code is heavily modified, I am wondering if the comment should then still be kept?

I added this comment to be able to differentiate between files that were never touched with AI tools from the ones that went through some kind of AI processing. This also allows us in the future to find the point where we added the first part of AI code and I think it is helpful to be transparent here.

matthias-ibach: Thanks for the explanation and your perspective on this topic!

I fully agree with you. Keeping this comment makes sense. Being able to distinguish the files that were never touched by AI from those that went through some kind of AI processing is definitely valuable for transparency. It also gives us a reference point for the future, for example if we ever need to trace back potential bugs, inconsistencies in style or other artifacts that might have been introduced through AI.

src/ebfm/core/cli.py
BenjaminRodenberg: This part caused quite some trouble because the code did not accept the `--version` command if the user did not pass some other sanity checks.

I assume that's the reason why the AI created such a lengthy comment here.

src/ebfm/core/cli.py
BenjaminRodenberg: The original motivation for providing default values centrally was to be able to use them in error/info messages without duplication.

I just moved them from `main.py` to `cli.py`. I think this place is ok. Or would you put the default somewhere else?

src/ebfm/core/cli.py
BenjaminRodenberg: I hadn't noticed this, but you're absolutely right. I was initially wondering whether the AI had silently changed the interface during the refactoring. However, we are indeed using `--couple-to-elmer-ice` in EBFM.

Apparently, this is a feature of `argparse` that I wasn't aware of. We could disable it by setting `allow_abbrev=False` on the parser, but that would be a breaking change because existing invocations may rely on the abbreviated form.

One argument for disabling abbreviations is forward compatibility. For example, today `ebfm --couple-to-icon` resolves unambiguously to `--couple-to-icon-atmo`. If a future release adds `--couple-to-icon-ocean`, the same invocation would become ambiguous and fail. Personally, I think we can live with this for now and revisit it if such a case arises. At least the failure mode is explicit—`argparse` raises an error rather than silently selecting the wrong option.

src/ebfm/core/cli.py
BenjaminRodenberg: Until now I did not experience a situation where I needed an entry in the README and the `--help` was not sufficient. You can, of course, always provide more detailed docs but I would only do this if there is a clear use-case / user-pain you are solving.

tox.ini
BenjaminRodenberg: Not sure, but let's try. If it is needed the corresponding tests should fail. I'll do a quick run.
```suggestion
```

src/ebfm/core/cli.py
BenjaminRodenberg: So to summarize: let's keep things as they are if you agree ;)

tox.ini
BenjaminRodenberg: https://github.com/EBFMorg/EBFM/pull/130/commits/f4872a4e4f212947caa2907844b6eb65c8ea875f fails. So apparently we need `dev` in the extras (pytest is missing, for example, if we skip `dev`).

BenjaminRodenberg: Reverted the change via e994c2591a97a5534298fb10b039550d98a1a61d

src/ebfm/core/cli.py
matthias-ibach: This is fine in my point of view, we can keep it here.

src/ebfm/core/cli.py
matthias-ibach: Yes, I agree!

I actually had the same train of thought, so we are on the same page here :-)

src/ebfm/core/cli.py
matthias-ibach: Yes, this makes sense, we can keep it as it is. And you are right, the `--help` is the best "first go to" (or second one if the README is not enough). Expanding the documentation only makes sense for repeated questions from users or clear use-cases.

tox.ini
matthias-ibach: I saw the failed tests, so apparently we need it. Let's keep it :-)

PR #135, opened 2026-06-25
Generalize file writing

### LINKED ISSUE #133: Generalize output options
Currently there is only an option for writing output if the grid type is GridInputType.MATLAB. It would be good to have some output functionality also for other, in particular irregular, grid types. The restart files could be used as an example. Some functionality to set output intervals and select variables would be good. Ideally, there would also be a time-averaging option.

wardvp:
### Implementation:

- [x] Update `LOOP_write_to_file.py` to include separate functions for writing to structured and unstructured NetCDF files.
- [x] Update `main.py` to write to file for all non-partitioned runs (coupled or uncoupled).
- [x] Update `config/grid.py` to set all non-MATLAB meshes to 'unstructured' as they all use the same Elmer mesh. **Previously `GridInputType.ELMER` and `GridInputType.CUSTOM` were classified as 'structured' but why? Please check and revert back when needed! I lost a bit track what the different GridInputTypes were used for...** 
- [x] Update INIT.py to fix missing conversion of latitude and longitude from radians to degrees when using the Elmer mesh.

### Tests:
- [x] Test output file writing (to `Output/model_output.nc`) for regular MATLAB grid (uncoupled run with random weather):
```python
ebfm --matlab-mesh examples/dem_and_mask.mat
``` 
- [x] Test output file writing (to `Output/model_output.nc`) for unstructured Elmer grid (uncoupled run with random weather):
```python
ebfm --elmer-mesh examples/greenland_mesh_v0/MESH --netcdf-mesh examples/BedMachineGreenland-v5_lo.nc --elmer-mesh-crs-epsg 3413 --with-numba --numba-threads 4
``` 
- [ ] Test output file writing (to `Output/model_output.nc`) for unstructured Elmer grid in a coupled run with ICON forcing. 
**I would be grateful if someone else tests this. I currently work in an EBFM version without coupling features.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,14 @@ Output/*.nc
 Reboot/*.nc
 .coverage
 .tox
+.idea/EBFM_with_YAC.iml
+.idea/inspectionProfiles/profiles_settings.xml
+.idea/misc.xml
+.idea/vcs.xml
+.idea/workspace.xml
+examples/1990_1.nc
+examples/1990_2.nc
+examples/add_mask_field_to_greenland_grid.py
+examples/createLowResGreenlandExperiment.py
+examples/download_CARRA2.py
+examples/preprocess_greenland_forcing_files.py
--- a/Output/Plot_variables_on_unstructured_grid.py
+++ b/Output/Plot_variables_on_unstructured_grid.py
@@ -1,0 +1,21 @@
+# SPDX-FileCopyrightText: 2025 EBFM Authors
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import netCDF4
+import matplotlib.pyplot as plt
+import matplotlib.tri as mtri
+
+nc = netCDF4.Dataset("model_output.nc")
+
+xv = nc["x_vertex"][:]
+yv = nc["y_vertex"][:]
+triangles = nc["cell_vertices"][:]
+smb = nc["smb"][0, :]
+
+triangulation = mtri.Triangulation(xv, yv, triangles)
+
+plt.tripcolor(triangulation, facecolors=smb, shading="flat")
+plt.colorbar(label="smb")
+plt.gca().set_aspect("equal")
+plt.show()
--- a/src/ebfm/core/config/grid.py
+++ b/src/ebfm/core/config/grid.py
@@ -85,7 +85,7 @@ class GridConfig:
             self.grid_type = GridInputType.CUSTOM
             self.mesh_file = elmer_mesh
             self.dem_file = args.netcdf_dem_mesh
-            self.is_unstructured = False
+            self.is_unstructured = True
         elif args.netcdf_dem_mesh_unstructured and elmer_mesh:
             self.grid_type = GridInputType.ELMERXIOS
             self.mesh_file = elmer_mesh
@@ -94,7 +94,7 @@ class GridConfig:
         elif elmer_mesh:
             self.grid_type = GridInputType.ELMER
             self.mesh_file = elmer_mesh
-            self.is_unstructured = False
+            self.is_unstructured = True
         else:
             logger.error(
                 f"Invalid grid configuration. EBFM supports the grid types {[t.name for t in GridInputType]}. "
--- a/src/ebfm/main.py
+++ b/src/ebfm/main.py
@@ -16,7 +16,6 @@ from ebfm.core import (
     LOOP_mass_balance,
 )
 from ebfm.core import LOOP_write_to_file, FINAL_create_restart_file
-from ebfm.core.grid import GridInputType
 from ebfm.core.config import CouplingConfig, GridConfig, TimeConfig
 from ebfm.core.logger import Logger, setup_logging, log_levels_map, getLogger
 from ebfm.core.cli import (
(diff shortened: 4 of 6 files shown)


.gitignore
wardvp, line 1: @BenjaminRodenberg I always need to add these local files in .gitignore to make sure I can commit my changes, but since these files are only on my local system I do not want them to appear in the general .gitignore. How to fix this?

src/ebfm/main.py
```diff
@@ -265,17 +264,11 @@ def _main_impl():
...
         # Write output to files (only in uncoupled run and for unpartitioned grid)
-        # TODO: should be supported for all cases to avoid case distinction here
-        if not grid["is_partitioned"] and isinstance(coupler, ebfm.coupling.DummyCoupler):
-            if grid_config.grid_type is GridInputType.MATLAB:
-                io, OUTFILE = LOOP_write_to_file.main(OUTFILE, io, OUT, grid, t, time)
-            else:
-                logger.warning("Skipping writing output to file for Elmer input grids.")
-        elif grid["is_partitioned"] or not isinstance(coupler, ebfm.coupling.DummyCoupler):
-            logger.warning("Skipping writing output to file for coupled or partitioned runs.")
+        # TODO: partitioned/coupled output should be supported in a dedicated parallel-safe writer.
```
wardvp, line 265: @BenjaminRodenberg Please check if this makes sense. File-writing would happen for all non-partitioned grids, regardless of whether the grids are structured or unstructured or whether coupled or uncoupled experiments are done.

src/ebfm/core/config/grid.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, args: Namespace):
             self.grid_type = GridInputType.CUSTOM
             self.mesh_file = elmer_mesh
             self.dem_file = args.netcdf_mesh
-            self.is_unstructured = False
+            self.is_unstructured = True
```
wardvp, line 88: @BenjaminRodenberg Please check. To me it seems that the same (unstructured) mesh (elmer_mesh) is used when GridInputType is CUSTOM, ELMER or ELMERXIOS, but maybe I am wrong...

BenjaminRodenberg: To me `self.is_unstructured = True` looks ok. Maybe this was an error previously. Will double check.

src/ebfm/main.py
BenjaminRodenberg: I will check this. We need to also support output for coupled runs but I cannot imagine a reason why it should not work for coupled runs if it does work for uncoupled runs (at least for the specific case we are interested in now)

.gitignore
BenjaminRodenberg: Sorry. This slipped my attention.

In `.git/info/exclude` you can add files that you want to ignore locally. Works identical to gitignore but is your private version that is not shared with the project team. I'm using the following:

```
EBFM_with_YAC$ cat .git/info/exclude 
# git ls-files --others --exclude-from=.git/info/exclude
# Lines that start with '#' are comments.
# For a project mostly in C, the following would be a good set of
# exclude patterns (uncomment them if you want to use them):
# *.[oa]
# *~

examples/elmer_w_DEM
examples/elmer_wo_DEM
examples/DEM
examples/BedMachine*

# local excludes

examples/1990_1_forcing_vectorized.nc
examples/1990_2_forcing_vectorized.nc
.venv*
*.orig
```

Important: If you delete your local copy of the repository the changes in this file will be gone because it's not tracked in git.

BenjaminRodenberg:
I rebased this branch onto `v0.7.0`. This now includes the crucial bugfix #149.

@hauke-schmidt reported that there is no regression compared to the status before the rebase.

BenjaminRodenberg:
I would like to merge this PR soon. Overall shape looks good to me and keeping it synchronized is causing a constant workload. I did a rebase and added some basic unit tests to make sure we at least cover the code (should avoid at least some mistakes during merging like I often do...). From the technical perspective it should be good to merge this PR.

There is an open todo "Test output file writing (to Output/model_output.nc) for unstructured Elmer grid in a coupled run with ICON forcing.". For now I added a warning to cover this gap. If this is something we can test quickly: Let's to it. If not: I would suggest to open an issue for this and merge. Here, I would need you judgement @hauke-schmidt / @wardvp on the open todo.

src/ebfm/main.py
```diff
@@ -180,6 +179,15 @@ def _main_impl():
...
     coupler.setup(grid, time_config)
 
+    # Warn once up front, rather than every timestep, about output that is written but not covered by tests.
+    # TODO: drop this once output writing is tested for an unstructured Elmer grid in a coupled run with ICON forcing.
+    is_coupled = not isinstance(coupler, ebfm.coupling.DummyCoupler)
+    if not grid["is_partitioned"] and (is_coupled or grid["is_unstructured"]):
+        logger.warning(
+            "Writing output to file is only tested for uncoupled runs on a structured grid; "
+            "check the contents of the output file."
+        )
```
BenjaminRodenberg: Does this cover the untested case?

Output/Plot_variables_on_unstructured_grid.py
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: 2025 EBFM Authors
```
BenjaminRodenberg:
```suggestion
# SPDX-FileCopyrightText: 2026 EBFM Authors
```

BenjaminRodenberg:
I cleaned this up and will merge it when the test are done.

I opened #176 to make sure we do not forget to run some more tests occasionally; if there are still bugs, we will find them for the features we are actually using sooner or later. For the features nobody uses we do not really need tests.

PR #141, opened 2026-06-30
Use ISO8601 as suffix for restart files

BenjaminRodenberg:
This PR provides an example on how to generate and use restart files with EBFM.

More important: It changes the formatting of file names to ISO8601 formatting. This can potentially break user's code but is consistent with new developments introduced in #137.

# Author's Todos


ClemensSchannwell:
We can introduce this, but what is the format of the time datetime object? Can you post an example of what the string looks like? I would want to avoid any spaces and characters like `+` ideally....

PR #143, opened 2026-07-07
Discontinue Python 3.10 support

BenjaminRodenberg:
Python 3.10 is reaching its [EOL in October 2026](https://devguide.python.org/versions/) and comes with some restrictions.

This PR updates the corresponding tests (drop 3.10 and test 3.14) and performs corresponding cleanup.

# Author's Todos

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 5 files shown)


matthias-ibach (commented):
Looks good.

As a side note, upgrading to a newer Python version on Levante for YAC is also mentioned [here](https://gitlab.dkrz.de/YAC/YAC-dev/-/work_items/291) and addressed [here](https://gitlab.dkrz.de/YAC/YAC-dev/-/merge_requests/602).

PR #144, opened 2026-07-09
Check `get_field_timestep` for `None`

BenjaminRodenberg:
Note: Added a guard in YAC via https://gitlab.dkrz.de/YAC/YAC-dev/-/commit/994af617020834736ea927ffdff63bc24107745b. Currently, a missing timestep leads to a segfault but this should be fixed in the next release and adding the guard here already now does not harm.

# Author's Todos


matthias-ibach (commented):
I guess the proper error handling happens in YAC? Or is it not necessary since `None` is returned in YAC?
Sorry for the questions, as I am not completely sure how this is handled.

PR #145, opened 2026-07-14
Explicitly check for Action.NO_DATA_RECEIVED

BenjaminRodenberg:
* Changed `CouplerErrorCode` to `CouplerExitCode` (not always error/fatal)
* Introduced check for action `NO_DATA_RECEIVED` to avoid getting and forwarding garbage values to user. Call `update` instead of `get` in this case.
* Improve logging.

# Author's Todos

### DIFF
--- a/src/ebfm/coupling/components/base.py
+++ b/src/ebfm/coupling/components/base.py
@@ -8,7 +8,6 @@ from collections.abc import Mapping, Callable
 import numpy as np
 
 from ebfm.core import logging
-from ebfm.coupling.couplers.base import CouplerExitCode
 
 logger = logging.getLogger(__name__)
 
@@ -102,16 +101,7 @@ class Component(ABC):
         data, err = self._coupler.get(self.name, field_name)
 
         if err:
-            if err is CouplerExitCode.NO_DATA_RECEIVED:
-                logger.debug(
-                    f"Get for {field_name=} returned exit code ({err=}). "
-                    "No data was received from the other component."
-                )
-            else:
-                logger.warning(
-                    f"Get for {field_name=} returned unexpected exit code ({err=}). "
-                    "Please report this to the developers."
-                )
+            logger.warning(f"Get for {field_name=} returned exit code ({err=}).")
             return None
 
         assert data is not None, f"Received data for field '{field_name}' is None. {err}"
--- a/src/ebfm/coupling/components/base.py
+++ b/src/ebfm/coupling/components/base.py
@@ -12,7 +12,7 @@ from ebfm.core import logging
 logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
-    from ebfm.coupling.couplers.base import Coupler
+    from ebfm.coupling.couplers.base import Coupler, CouplerExitCode
     from ebfm.coupling.fields.base import FieldSet
 
 
@@ -79,7 +79,8 @@ class Component(ABC):
             err = self._coupler.put(self.name, field_name, transform(data_to_exchange[field_name]))
             if err:
                 logger.warning(
-                    f"Put for {field_name=} returned error code ({err=}). Please report this to the developers."
+                    f"Put for {field_name=} returned unexpected exit code ({err=}). "
+                    "Please report this to the developers."
                 )
 
     def _get_if_coupled(self, field_name: str, transform: Callable = identity) -> np.ndarray | None:
@@ -96,9 +97,16 @@ class Component(ABC):
             data, err = self._coupler.get(self.name, field_name)
             logger.debug(f"Received data for field '{field_name}' from coupler: {data}")
             if err:
-                logger.warning(
-                    f"Get for {field_name=} returned error code ({err=}). Please report this to the developers."
-                )
+                if err is CouplerExitCode.NO_DATA_RECEIVED:
+                    logger.debug(
+                        f"Get for {field_name=} returned exit code ({err=}). "
+                        "No data was received from the other component."
+                    )
+                else:
+                    logger.warning(
+                        f"Get for {field_name=} returned unexpected exit code ({err=}). "
+                        "Please report this to the developers."
+                    )
             assert data is not None, f"Received data for field '{field_name}' is None. {err}"
             return transform(data)
         return None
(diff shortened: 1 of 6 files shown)


src/ebfm/coupling/couplers/yacCoupler.py
```diff
@@ -243,9 +240,25 @@ def get(self, component_name: str, field_name: str) -> tuple[np.ndarray | None,
...
+            field.field_handle.update()
+            logger.debug(f"No data for field {field.name} from {field.coupled_component.name} received.")
+            return None, CouplerExitCode.NO_DATA_RECEIVED
+
+        expected_actions = {yac.Action.GET, yac.Action.GET_FOR_RESTART}
+
+        assert action in expected_actions, (
+            f"Expected action {expected_actions} for field '{field.name}' from component "
+            f"'{field.coupled_component.name}', but got {action}."
+        )
```
BenjaminRodenberg, line 270: Would you rather check explicitly for `if action is yac.Action.NONE` or `if action not in expected_actions`?

With the assertion it should be safe, but I do not know if `NONE` catches all relevant cases?

I think the current approach is the most explicit one; therefore, I picked that one. With the assertion we will at least know if something unexpected happens here.

src/ebfm/coupling/couplers/yacCoupler.py
```diff
@@ -194,25 +194,22 @@ def put(self, component_name: str, field_name: str, data: np.ndarray) -> Coupler
...
         Get data from another component
 
         @param[in] component_name name of the component to get data from
         @param[in] field_name name of the field to get data for
 
-        @returns tuple of (field data, error code). Error code is None if no error occurred.
-                 Field data is always the value returned by YAC (e.g. zeros as fallback) even
-                 when an error code is set, so the caller can decide whether to use it or substitute
-                 their own fallback.
+        @returns tuple of (field data, exit code). Exit code is None if get successfully received data.
```
BenjaminRodenberg, line 212: @matthias-ibach we do not return all data received from YAC anymore. Returning garbage values is probably a bad idea so I think it's fine to break this contract with the calling functions and change the behavior.

I assume this does not cause problems, but I vaguely remember that we introduced this together in https://github.com/EBFMorg/EBFM/pull/87 for some reason...

src/ebfm/coupling/components/base.py
```diff
@@ -96,9 +97,16 @@ def _get_if_coupled(self, field_name: str, transform: Callable = identity) -> np
...
+                        "No data was received from the other component."
+                    )
+                else:
+                    logger.warning(
+                        f"Get for {field_name=} returned unexpected exit code ({err=}). "
+                        "Please report this to the developers."
+                    )
             assert data is not None, f"Received data for field '{field_name}' is None. {err}"
             return transform(data)
         return None
```
k202077: How about:
```
if not_has_field return None
temp_data, err = get
assert (err == none) || (err is NO_DATA)
data = none
if err
  logger
else
  logger
  data = transform(temp_data)
return data
```

src/ebfm/coupling/couplers/yacCoupler.py
```diff
@@ -243,9 +240,25 @@ def get(self, component_name: str, field_name: str) -> tuple[np.ndarray | None,
...
+            logger.debug(f"No data for field {field.name} from {field.coupled_component.name} received.")
+            return None, CouplerExitCode.NO_DATA_RECEIVED
+
+        expected_actions = {yac.Action.GET, yac.Action.GET_FOR_RESTART}
+
+        assert action in expected_actions, (
+            f"Expected action {expected_actions} for field '{field.name}' from component "
+            f"'{field.coupled_component.name}', but got {action}."
+        )
+
```
k202077: There is also `OUT_OF_BOUNDS`.  
This usually occurs in case of a configuration error (number of puts/gets for a coupled field do not match between two components).  
Add dedicated error/warning message (warnings are usually ignored...).

src/ebfm/coupling/couplers/yacCoupler.py
k202077: Looks good.  
`OUT_OF_BOUND` could potentially get a dedicated handling (see comment below).

src/ebfm/coupling/couplers/yacCoupler.py
BenjaminRodenberg: If we want to treat this as a fatal error (sounds to me like we want to) the assertion should be sufficient. We could also raise a more specific error.

How does this look like?

```suggestion
        action = field.field_handle.action()

        if action is yac.Action.NONE:
            logger.debug(f"Received YAC action NONE for field {field.name} from {field.coupled_component.name}.")
            logger.debug(f"Skipping get and updating field {field.name} to next time step.")
            field.field_handle.update()
            logger.debug(f"No data for field {field.name} from {field.coupled_component.name} received.")
            return None, CouplerExitCode.NO_DATA_RECEIVED

        if action is yac.Action.OUT_OF_BOUNDS:
            logger.warning(f"Received YAC action OUT_OF_BOUNDS for field {field.name} "
                         "from {field.coupled_component.name}.")
            logger.warning(f"Skipping get and updating field {field.name} to next time step.")
            field.field_handle.update()
            logger.warning(f"Please review your coupling configuration for errors.")
            logger.debug(f"No data for field {field.name} from {field.coupled_component.name} received.")
            return None, CouplerExitCode.NO_DATA_RECEIVED
            
        expected_actions = {yac.Action.GET, yac.Action.GET_FOR_RESTART}

        assert action in expected_actions, (
            f"Expected action {expected_actions} for field '{field.name}' from component "
            f"'{field.coupled_component.name}', but got {action}."
        )
```

I assume `yac.Action.OUT_OF_BOUNDS` leads to `NO_DATA_RECEIVED`? We could also make the whole thing fatal if you prefer.

src/ebfm/coupling/components/base.py
BenjaminRodenberg: Rearranged it in 4665ea4ad6b1a80493f023d948db537584278a00.

src/ebfm/coupling/couplers/yacCoupler.py
k202077: yes, in case of `OUT_OF_BOUNDS` no data will be changed

Warning or error is up to you.

Looks like you could use `match` instead of multiple `if` and add a `raise` in the default.

PR #146, opened 2026-07-14
Add general interface for definition of fallback

BenjaminRodenberg:
# Author's Todos

### DIFF
--- a/src/ebfm/coupling/couplers/base.py
+++ b/src/ebfm/coupling/couplers/base.py
@@ -78,7 +78,7 @@ class Coupler(ABC, Generic[CouplerExchangeType]):
 
         logger.debug(f"Active coupled components: {list(self._coupled_components.keys())}")
 
-        self.fields: FieldSet = FieldSet()
+        self._fields: FieldSet = FieldSet()
         self._time: TimeConfig | None = None  # will be set in setup()
 
     @staticmethod
@@ -214,7 +214,7 @@ class Coupler(ABC, Generic[CouplerExchangeType]):
 
         expected_exchange_type = self._map_exchange_type(exchange_type)
         component = self._coupled_components[component_name]
-        fields = self.fields.filter(
+        fields = self._fields.filter(
             lambda f: f.coupled_component == component
             and f.name == field_name
             and f.exchange_type == expected_exchange_type
--- a/src/ebfm/coupling/couplers/base.py
+++ b/src/ebfm/coupling/couplers/base.py
@@ -31,11 +31,13 @@ Examples:
 """
 
 
-class CouplerErrorCode(Enum):
+class CouplerExitCode(Enum):
     """
-    Error codes returned by Coupler.get() and Coupler.put().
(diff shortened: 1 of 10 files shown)


src/ebfm/main.py
```diff
@@ -193,7 +193,15 @@ def _main_impl():
                 "albedo": OUT["albedo"],
             }
 
-            data_from_icon = icon_atmo.exchange(data_to_icon)
+            fallback_values = {
+                "rlds": IN["LWin"],
+                "clt": IN["C"],
+                "sfcwind": IN["WS"],
+                "huss": IN["T"] * 0.0,
+                "sfcpres": IN["T"] * 0.0 + 101500.0,
```
matthias-ibach, line 201: Should the fallback values only contain these variables? What about `pr, pr_snow, rsds, tas`? 
But, maybe I don't understand the exchange between ICON and EBFM entirely.

matthias-ibach (commented):
Otherwise, this looks good to me. 

I tested this version with our TerraDT testcase and it worked., see example below:

```
...
EBFM 2026-07-15 09:18:05.198 WARNING base - _get_if_coupled: Get for field_name='rlds' returned exit code (err=<CouplerExitCode.NO_DATA_RECEIVED: 'no_data_received'>).
EBFM 2026-07-15 09:18:05.198 WARNING base - _get_if_coupled: Using fallback value for 'rlds' as no data was received for this field from icon_atmo.
...
```

src/ebfm/main.py
BenjaminRodenberg: I would like to keep the definition of fallbacks as focused as possible for now. The risk with fallbacks is that they are silently used without the user wanting it. So I personally prefer an error here. With the interface it should be relatively easy to add new fallbacks if needed.

Maybe there should also be an option such that the user has to explicitly allow for the use of any fallbacks with the CLI. But for now I think it's ok as it is with the warning.

Hope this strategy makes sense?

matthias-ibach: Yes, that sounds very reasonable!

tests/coupling/test_components.py
BenjaminRodenberg, line 1: Added this test to be able to detect the bug which required the fix https://github.com/EBFMorg/EBFM/commit/91edef5c986ef04f53e5d8fceb515671b0b53dc3.

src/ebfm/coupling/couplers/base.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, coupling_config: CouplingConfig):
 
         logger.debug(f"Active coupled components: {list(self._coupled_components.keys())}")
 
-        self.fields: FieldSet = FieldSet()
+        self._fields: FieldSet = FieldSet()
```
BenjaminRodenberg, line 81: Changed this to `_fields` to indicate it's an internal field that a user should normally not access. I was a bit careless in the test and forgot to call `_register_fake_values`. I hope this is now clearer by declaring it as internal.

PR #147, opened 2026-07-28
Further CPU (host-side) performance optimization (in LOOP_SNOW.py)

matthias-ibach:
This MR introduced further optimizations on the host-side/CPU (non-kernel) parts of `LOOP_SNOW`. No physics changes.
It additionally serves as preparation for GPU offloading capability of EBFM.

@wardvp : You can check your actual production cases using this branch to test if your results remain unchanged. These performance optimizations should further speed-up your runtime up to 1.5x (see below).
It is important though to check whether your more realistic cases produce the same results.

### What changed

* `layer_merging_and_splitting`, `snowfall_and_deposition`, `melt_sublimation`: snapshot/copy only the rows/columns that are actually updated, instead of copying the full `(gpsum, nl)` grid on every step
* `compaction`, `percolation`: the `*_old` snapshots are taken per column inside the Numba kernels instead of on the host
* `heat_conduction`: the precompute (`kk`, `dz1/dz2`, `denom_*`, `dt_stab`) moved into a new `@njit` kernel `_heat_conduction_prep_kernel`. The NumPy path remains unchanged.
* Added `@profile` to the remaining `LOOP_SNOW` functions so `kernprof` covers all of them.
* The `performance` extra no longer installs `mpi4py` (it was never necessary here: Numba parallelises with threads inside a single process and does not need MPI).

## Improvement

On Levante (similar numbers on local copmuter) using the Elmer mesh example (gpsum 372k):

| | baseline | now | |
|---|---|---|---|
| `LOOP_SNOW` (numba) | 21.0 s | 7.0 s | **3.0x** |
| total runtime (numba) | 45.3 s | 31.7 s | **1.43x** |
| total runtime (numpy) | 135 s | 129 s | 1.05x |

The NumPy path gains little because its cost is the CFL heat solver and the percolation physics, which are untouched.

Results are bit-identical (`max |diff| = 0`, via `--dump-reference` + `tools/compare_snapshots.py`) for every commit, on both the matlab and Elmer mesh and with both backends.

# Author's Todos

### DIFF
--- a/src/ebfm/core/LOOP_SNOW.py
+++ b/src/ebfm/core/LOOP_SNOW.py
@@ -6,7 +6,12 @@ import numpy as np
 
 from .compute_backend import get_backend, ComputeBackend
 from .constants import SECONDS_PER_HOUR
-from .LOOP_SNOW_kernels import _compaction_kernel, _heat_conduction_kernel, _percolation_kernel
+from .LOOP_SNOW_kernels import (
+    _compaction_kernel,
+    _heat_conduction_kernel,
+    _heat_conduction_prep_kernel,
+    _percolation_kernel,
+)
 
 # line_profiler support: `profile` is injected as a builtin by kernprof.
 # When running normally, fall back to a no-op so the decorator stays in place.
@@ -415,40 +420,39 @@ def main(C, OUT, IN, dt, grid, phys):
         """
         Calculate heat diffusion and update temperatures
         """
-        dz1 = (OUT["subZ"][:, 0] + 0.5 * OUT["subZ"][:, 1]) ** 2
-        dz2 = 0.5 * (OUT["subZ"][:, 2:] + OUT["subZ"][:, 1:-1]) ** 2
-        kk = 0.138 - 1.01e-3 * OUT["subD"] + 3.233e-6 * OUT["subD"] ** 2  # Effective conductivity
-        c_eff = OUT["subD"] * (152.2 + 7.122 * OUT["subT"])  # Effective heat capacity
-
-        # Stability time step (CFL condition)
-        # Layer 0: surface ghost layer, excluded from CFL condition
-        dt_stab = (
-            0.5
-            * np.min(c_eff[:, 1:], axis=1)
-            * np.min(OUT["subZ"][:, 1:], axis=1) ** 2
-            / np.max(kk[:, 1:], axis=1)
-            / C["dayseconds"]
-        )
-        assert (dt_stab > 0).all(), "cells with dt_stab <= 0 are forbidden!"
-
-        # subZ and c_eff do not change
-        # Precompute kk*subZ products once
-        # kk_sz_top: conductivity-thickness product for the top interface
-        # kk_sz_interior: same for all interior interfaces
-        kk_sz_top = kk[:, 0] * OUT["subZ"][:, 0] + 0.5 * kk[:, 1] * OUT["subZ"][:, 1]
-        kk_sz_interior = kk[:, 1:-1] * OUT["subZ"][:, 1:-1] + kk[:, 2:] * OUT["subZ"][:, 2:]
-
-        # Precompute full temperature-update denominators once
-        # denom_layer1: first active layer (layer 0: surface ghost layer overwritten from Tsurf)
-        denom_layer1 = c_eff[:, 1] * (0.5 * OUT["subZ"][:, 0] + 0.5 * OUT["subZ"][:, 1] + 0.25 * OUT["subZ"][:, 2])
-        denom_interior = c_eff[:, 2:-1] * (
-            0.25 * OUT["subZ"][:, 1:-2] + 0.5 * OUT["subZ"][:, 2:-1] + 0.25 * OUT["subZ"][:, 3:]
-        )
-        denom_bottom = c_eff[:, -1] * (0.25 * OUT["subZ"][:, -2] + 0.75 * OUT["subZ"][:, -1])
-
         # ------ Heat Conduction Loop ------
         if get_backend() == ComputeBackend.NUMBA:
-            # Numba parallel path: prange(gpsum), each column solved independently.
+            # Numba parallel path: the per-column precompute that the NumPy branch
+            # below does in host NumPy is done in _heat_conduction_prep_kernel
+            # (prange over gpsum); the CFL solve then runs in _heat_conduction_kernel.
+            gpsum, nl = OUT["subT"].shape
+            kk = np.empty((gpsum, nl))
+            c_eff = np.empty((gpsum, nl))
+            kk_sz_top = np.empty(gpsum)
+            kk_sz_interior = np.empty((gpsum, nl - 2))
+            dz1 = np.empty(gpsum)
+            dz2 = np.empty((gpsum, nl - 2))
+            denom_layer1 = np.empty(gpsum)
+            denom_interior = np.empty((gpsum, nl - 3))
+            denom_bottom = np.empty(gpsum)
+            dt_stab = np.empty(gpsum)
+            _heat_conduction_prep_kernel(
+                OUT["subD"],
+                OUT["subZ"],
+                OUT["subT"],
+                kk,
+                c_eff,
+                kk_sz_top,
+                kk_sz_interior,
+                dz1,
+                dz2,
+                denom_layer1,
+                denom_interior,
+                denom_bottom,
+                dt_stab,
+                C["dayseconds"],
+            )
+            assert (dt_stab > 0).all(), "cells with dt_stab <= 0 are forbidden!"
             _heat_conduction_kernel(
                 OUT["subT"],
                 OUT["Tsurf"],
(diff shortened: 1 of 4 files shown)


src/ebfm/core/LOOP_SNOW.py
```diff
@@ -80,58 +86,62 @@ def snowfall_and_deposition():
...
 
-            # Use references instead of unnecessary .copy()
-            subT_old = OUT["subT"].copy()
-            subD_old = OUT["subD"].copy()
-            subW_old = OUT["subW"].copy()
-            subZ_old = OUT["subZ"].copy()
+            # No-shift branch only touches top layer, only column 0 copied (instead of whole grid).
+            subZ0_old = OUT["subZ"][:, 0].copy()
+            subT0_old = OUT["subT"][:, 0].copy()
+            subD0_old = OUT["subD"][:, 0].copy()
```
BenjaminRodenberg:
```suggestion
            # No-shift branch only touches top layer, only column 0 copied (instead of whole grid).
            subZ_top_old = OUT["subZ"][:, 0].copy()
            subT_top_old = OUT["subT"][:, 0].copy()
            subD_top_old = OUT["subD"][:, 0].copy()
```
A bit more descriptive and less implementation specific; but I can also live with `subZ0_old` etc.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -80,58 +86,62 @@ def snowfall_and_deposition():
...
 
             # Precompute conditions for better performance
-            is_noshift = subZ_old[:, 0] + shift <= max_subZ
+            is_noshift = subZ0_old + shift <= max_subZ
             is_shift = ~is_noshift
 
             # Handle no-shift updates (vectorized)
             OUT["subZ"][is_noshift, 0] += shift[is_noshift]
+            z0_new = OUT["subZ"][is_noshift, 0]
```
BenjaminRodenberg: Is there a difference between `z0` and `subZ0`? If not, use the following:
```suggestion
            subZ0_new = OUT["subZ"][is_noshift, 0]
```

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -166,45 +176,51 @@ def melt_sublimation():
...
+            subD0_old = OUT["subD"][:, 0].copy()
+            subW0_old = OUT["subW"][:, 0].copy()
 
-            # Handle the no-shift case
-            OUT["subZ"][i_noshift, 0] = subZ_old[i_noshift, 0] + shift[i_noshift]
-            OUT["subT"][i_noshift, 0] = subT_old[i_noshift, 0]
-            OUT["subD"][i_noshift, 0] = subD_old[i_noshift, 0]
-            temp = OUT["subZ"][i_noshift, 0] / subZ_old[i_noshift, 0]
-            OUT["subW"][i_noshift, 0] = subW_old[i_noshift, 0] * temp
+            noshift = subZ0_old + shift > 1e-17
```
BenjaminRodenberg: The `shift > 1e-17` looks a bit magic to me.

BenjaminRodenberg: Additionally: What does `noshift` actually mean? It's a bit hard to understand what's going on here.

The `shift > 1e-17` looks like it is a boolean array; maybe use `or` instead of the `+`?

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -166,45 +176,51 @@ def melt_sublimation():
...
+            idx_shift = np.flatnonzero(~noshift)
```
BenjaminRodenberg:
```suggestion
            needs_shift = ~noshift
            idx_shift = np.flatnonzero(needs_shift)
```
A double negation looks a bit unnecessary, but since you are using `noshift` below maybe like this?

I also do not really like the `~` because it's easy to mix up with `-` (but maybe this is my fault or the fault of my editor). Again: I think explicitly declaring the `dtype` here with a type annotation would help understanding that we are dealing with booleans here.

src/ebfm/core/LOOP_SNOW.py
BenjaminRodenberg: Using a boolean `noshift` and a non boolean `shift` is hard to understand. What does `shift` store? Maybe `shift_amount` if this fits. Currently one could assume that `shift == ~noshift`.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -166,45 +176,51 @@ def melt_sublimation():
...
+            OUT["subD"][noshift, 0] = subD0_old[noshift]
+            temp = z0_new / subZ0_old[noshift]
+            OUT["subW"][noshift, 0] = subW0_old[noshift] * temp
+
+            # Handle the shift case: copy only the shift rows.
+            if idx_shift.size:
+                subZ_old = OUT["subZ"][idx_shift]
+                subT_old = OUT["subT"][idx_shift]
+                subD_old = OUT["subD"][idx_shift]
+                subW_old = OUT["subW"][idx_shift]
```
BenjaminRodenberg, line 232: Using `idx_shift` for the cases where a shift is needed and a boolean mask `noshift` for the cases where no shift is needed is confusing. I would stick to one or the other approach and not mix them here.

See above: If you use `needs_shift` you should be able to use `subZ_old = OUT["subZ"][needs_shift]` and you to not have to generate the indices.

`idx_shift.size` would become `needs_shift.any()`.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -278,6 +288,9 @@ def compaction():
             )
         else:
             # NumPy path
+            subD_old = OUT["subD"].copy()
+            subZ_old = OUT["subZ"].copy()
+            mliqmax = np.zeros((gpsum, nl))
```
BenjaminRodenberg: `mliqmax` needs a comment describing what this means or a better name

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -463,7 +474,35 @@ def heat_conduction():
...
+            # NumPy path: host precompute + explicit while-loop with vectorized
+            # column updates.
+            dz1 = (OUT["subZ"][:, 0] + 0.5 * OUT["subZ"][:, 1]) ** 2
+            dz2 = 0.5 * (OUT["subZ"][:, 2:] + OUT["subZ"][:, 1:-1]) ** 2
+            kk = 0.138 - 1.01e-3 * OUT["subD"] + 3.233e-6 * OUT["subD"] ** 2  # Effective conductivity
+            c_eff = OUT["subD"] * (152.2 + 7.122 * OUT["subT"])  # Effective heat capacity
+
+            # Stability time step (CFL condition)
+            # Layer 0: surface ghost layer, excluded from CFL condition
+            dt_stab = (
```
BenjaminRodenberg:
```suggestion
            dt_max = (
```
?

src/ebfm/core/LOOP_SNOW.py
BenjaminRodenberg: Initializing `tt` with an array looks wrong to me and makes the `while...` below confusing. I'll open an independent PR for this because this is out-of-scope here.

BenjaminRodenberg (commented):
A few comment on readability and naming.

src/ebfm/core/LOOP_SNOW.py
BenjaminRodenberg: I was totally unaware of the fact that the solver performs local time stepping and this was also not obvious in the code. Here is the PR: https://github.com/EBFMorg/EBFM/pull/153. Hope this clarifies the situation.

src/ebfm/core/LOOP_SNOW.py
BenjaminRodenberg: See https://github.com/EBFMorg/EBFM/pull/153. Don't worry here.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: Yes, I agree. I have changed it.
`subZ_top_old/subT_top_old/subD_top_old/subW_top_old` is now present in both loops.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: Done, I changed this. Now, we have `subZ_top_new`. Additionally, the `temp` intermediates are gone too.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: After consulting Claude, it could actually exactly tell me the history of this "magic number", roughly stating:

"`1e-17` is not new. It comes from the initial commit (of EBFM) (`42c0113, src/ebfm/LOOP_SNOW.py:150`) and is unchanged on main (`LOOP_SNOW.py:176-177`, inside np.where).  this MR only folded the two np.where calls into one boolean mask. It's a floating-point zero guard on the remaining top-layer thickness (exact 0.0 would let a ~1e-20 m layer survive), now named `_MIN_LAYER_THICKNESS` with a comment."

Additionally, `shift` is a thickness in metres, not a boolean, so the `+` is arithmetic. I went ahead and renamed it to `shift_amount`, with `noshift` meaning `no_grid_shift` plus a short comment on what shifting the grid means.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: You are right. I did a change following `shift` to  `shift_amount` (pairs with `shift_tot`) and the masks now say `grid_shift` so they cannot be read as the amount.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: Thanks. Now, the `~` is down to a single occurrence on its own named line (`needs_grid_shift = ~no_grid_shift`  as you suggested), plus a trailing `# bool per grid point` (comment) on the mask itself.

On the type annotation: I agree in general. But so far, `LOOP_SNOW.py` has no annotations at all right now, so the two `npt.NDArray[np.bool_]` here would be the only typed lines in the module. I think I would rather not start that in this (performance) MR but address this in a separate one.

But if you think it makes sense to address it here, I will be happy to do so.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: You are correct with regards to "not mixing, but unified on index arrays rather than masks".

I fixed it, but the other way round: both branches now use index arrays (`idx_noshift/idx_shift`) rather than both using masks. Reason (which I found out when consulting AI): NumPy re-scans a boolean mask across the whole grid on every indexing operation and the shift branch does ~14 of them while touching only a handful of rows. `np.flatnonzero` pays that scan once and each later access then costs only the affected rows. On the elmer mesh (`gpsum 372k`, ~0.1 % of points shifting) that's 0.6 ms vs 19.3 ms for the shift branch alone. It also matches `idx_merge/idx_split` in `layer_merging_and_splitting`, so the file is consistent now. The condition itself is still a named boolean (`no_grid_shift/needs_grid_shift`), just converted to indices once.

If you think having it consistent using masks though is the better way, I can do that as well. The cost is really small in an actual production run.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: I added a comment (max liquid water a layer holds against gravity, kg m-2) and moved the allocation to appear where it is used. I kept the name since the kernels and `percolation_refreezing_and_storage` use it too.

src/ebfm/core/LOOP_SNOW.py
BenjaminRodenberg: Both approaches are fine and cost saving are a good reason to prefer indices over masks.

Not sure if this is common knowledge or something worth a comment in the code?

src/ebfm/core/LOOP_SNOW.py
BenjaminRodenberg: The nice thing about type annotations is in my opinion that you can opt-in incrementally. There is no need to consequently use type annotations in the whole file or code base. So I would add the annotations now for the cases where they improve readability. But still: It is optional and if you prefer concentrating on performance that's also fine.

matthias-ibach: Yes, that is actually a valid and good point on incremental opt-in.  
I went ahead and `NDArray[np.bool_]` on the two masks and `NDArray[np.intp]` on the index arrays, since that mask-vs-index distinction is what was hard to read here.

src/ebfm/core/LOOP_SNOW.py
matthias-ibach: You are definitely right that the reasoning here is not obvious, it actually involved quite a bit of research, verification and arguing with AI.  
So I went ahead and added a short comment:  
"NumPy re-scans the full mask on every indexing operation, while flatnonzero pays that scan once and each later access then costs only the affected rows."

PR #153, opened 2026-08-10
Revise local time stepping

BenjaminRodenberg:
Added some comments and improved readability.

Motivation: It was quite hard to understand that columns use local time stepping.

# Author's Todos

### DIFF
--- a/src/ebfm/core/LOOP_SNOW.py
+++ b/src/ebfm/core/LOOP_SNOW.py
@@ -493,7 +493,8 @@ def main(C, OUT, IN, dt: float, grid, phys):
 
                 # Integer indices of still-active columns
                 idx = np.flatnonzero(dt_local > 0)
-                assert idx.size > 0, "no cells left where local time stepping has to be performed. Unexpected here."
+                # See https://github.com/EBFMorg/EBFM/pull/154
+                # assert idx.size > 0, "no cells left where local time stepping has to be performed. Unexpected!"
 
                 kdTdz[idx, 1] = kk_sz_top[idx] * (T_old[idx, 1] - OUT["Tsurf"][idx]) / dz1[idx]
                 kdTdz[idx, 2:] = kk_sz_interior[idx] * (T_old[idx, 2:] - T_old[idx, 1:-1]) / dz2[idx]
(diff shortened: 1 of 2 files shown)


src/ebfm/core/LOOP_SNOW.py
```diff
@@ -418,15 +419,17 @@ def heat_conduction():
...
         # Layer 0: surface ghost layer, excluded from CFL condition
-        dt_stab = (
+        dt_stab: NDArray[np.float64] = (
             0.5
             * np.min(c_eff[:, 1:], axis=1)
             * np.min(OUT["subZ"][:, 1:], axis=1) ** 2
             / np.max(kk[:, 1:], axis=1)
             / C["dayseconds"]
         )
+        assert (dt_stab > 0).all(), "cells with dt_stab <= 0 are forbidden!"
```
BenjaminRodenberg, line 443:
```suggestion
        # See https://github.com/EBFMorg/EBFM/pull/154
        # assert (dt_stab > 0).all(), "cells with dt_stab <= 0 are forbidden!"
```
Fails for the test case with restart. I would treat this as an (independent) bug to continue here and resolve the bug independently.

src/ebfm/core/LOOP_SNOW.py
```diff
@@ -463,31 +467,39 @@ def heat_conduction():
...
-                    break
-                # Calculate vertical heat fluxes
+
+                dt_remainder = dt - t_elapsed  # determine remainder until end of global time step
+                # use minimum of 1) maximum allowed dt due to stability and 2) remainder until end of global time step
+                dt_local = np.minimum(dt_stab, dt_remainder)
+
+                # Integer indices of still-active columns
+                idx = np.flatnonzero(dt_local > 0)
+                assert idx.size > 0, "no cells left where local time stepping has to be performed. Unexpected here."
```
BenjaminRodenberg:
```suggestion
                # See https://github.com/EBFMorg/EBFM/pull/154
                # assert idx.size > 0, "no cells left where local time stepping has to be performed. Unexpected!"
```

matthias-ibach (commented):
This looks good to me, it makes the local time stepping clearer.

PR #154, opened 2026-08-10
Ensure `(dt_stab > 0).all()`

BenjaminRodenberg:
# Author's Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Fixed bug where resuming from a restart file (`--restart-init`) could lead to time step sizes equal to zero in `LOOP_SNOW.py`. Restart loading now asserts restart variables contain no missing values and converts them to plain `ndarray`. https://github.com/EBFMorg/EBFM/pull/154
 * Fixed bug in icon_to_atmo.py to convert pr_snow units to mwe per timestep (rather than using kg m-2 s-1). https://github.com/EBFMorg/EBFM/pull/149
 * Add general interface for definition of fallback values if a coupled component does not provide expected data for individual fields. https://github.com/EBFMorg/EBFM/pull/146.
     * The following fields expected from the component "icon_atmo" will use fallback value if they are missing: "rlds", "clt", "sfcwind", "huss", and "sfcpres"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
-* Fixed bug where resuming from a restart file (`--restart-init`) could lead to time step sizes equal to zero in `LOOP_SNOW.py`. Restart loading now asserts restart variables contain no missing values and converts them to plain `ndarray`. https://github.com/EBFMorg/EBFM/pull/154
+* Fixed bug where resuming from a restart file (`--restart-init`) could lead to time step sizes equal to zero in `LOOP_SNOW.py`. Loading a restart file now asserts the restart variables to contain no missing values and converts them to plain `ndarray`. https://github.com/EBFMorg/EBFM/pull/154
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Fixed bug where resuming from a restart file (`--restart-init`) could lead to time step sizes equal to zero in `LOOP_SNOW.py`. Restart loading now asserts restart variables contain no missing values and converts them to plain `ndarray`. https://github.com/EBFMorg/EBFM/pull/154
```
matthias-ibach: I kept reading the second sentence, I needed some tries to understand it. Maybe rephrase to `Loading a restart file now asserts the restart variables to contain no missing values and...`. Might not be the best English, but was easier for me to undertstand.

src/ebfm/core/INIT.py
```diff
@@ -554,8 +554,22 @@ def init_initial_conditions(C, grid: GridDict, io, time, init_with_restart_file:
...
-                # Read the variable data
+                # Read the variable data. netCDF4 returns a numpy.ma.MaskedArray
+                # here whenever the variable defines a fill value, regardless of
+                # whether any missing values are actually present.
                 var_data = ncfile.variables[var_name][:]
+                # Restart files are expected to be fully populated. Mixing masked
+                # arrays into the simulation state is unsafe: numpy's masked-array
+                # ufuncs silently ignore `where=` when combining masks (only the
+                # data respects `where`), so masks can leak into cells they
+                # shouldn't and snowball across timesteps. Verify the assumption
```
matthias-ibach: I usually like `should not` instead of `shouldn't` in "official" documentation, but that is just personal preference.

src/ebfm/core/INIT.py
```diff
@@ -554,8 +554,22 @@ def init_initial_conditions(C, grid: GridDict, io, time, init_with_restart_file:
...
+                # holds, then drop down to a plain ndarray.
+                assert not np.ma.is_masked(var_data), (
```
matthias-ibach, line 567: So if I understood correctly, there is no other way other than to abort when there are missing values in the restart file? So there is no mechanism in Python regarding masks to support this? Sorry if I do not quite get this right.

tests/core/test_init.py
```diff
@@ -0,0 +1,115 @@
...
+
+class TestInitFromRestartFile(unittest.TestCase):
+    """
+    Regression test for a bug where restart-loaded arrays stayed
+    `numpy.ma.MaskedArray` (netCDF4's default) instead of plain `ndarray`.
+
+    Mixing masked arrays into the simulation state made numpy's masked-array
+    ufuncs silently ignore `where=` when combining masks in LOOP_SNOW's
+    compaction() step, so masks from expected divide-by-zero results leaked
+    into cells they shouldn't and snowballed across timesteps, eventually
```
matthias-ibach:
```suggestion
    into cells they should not and snowballed across timesteps, eventually
```

matthias-ibach (commented):
Looks good to me, just some minor comments regarding style and wording.

src/ebfm/core/INIT.py
BenjaminRodenberg: I would assume that restart files have complete data. If this assumption is not correct, fail loudly instead of continue silently.

src/ebfm/core/INIT.py
BenjaminRodenberg:
```suggestion
                # should not and snowball across timesteps. Verify the assumption
```

CHANGELOG.md
BenjaminRodenberg:
```suggestion
* Fixed bug where resuming from a restart file (`--restart-init`) could lead to time step sizes equal to zero in `LOOP_SNOW.py`. Loading a restart file now asserts the restart variables to contain no missing values and converts them to plain `ndarray`. https://github.com/EBFMorg/EBFM/pull/154
```

PR #155, opened 2026-08-11
Skip the shading look-up table when shading is disabled

matthias-ibach:
For MATLAB grids, `init_grid` always pre-computed the maximum-grid-angle look-up table (`grid["maxgridangle"]`), even when shading was switched off.

That table is only read by the shading branch of `LOOP_EBM_insolation`, which is itself guarded by `grid["has_shading"]`. With shading disabled the whole loop ran at startup and the result was discarded.

The fix guards the pre-computation with `grid["has_shading"]`. With shading enabled nothing changes.

On the Svalbard example grid (`examples/dem_and_mask.mat`, 37993 glacier cells) the costs were:

shading | init_grid
-- | --
off (before) | 8.2 s
off (after) | 0.07 s
on | 8.2 s, unchanged

The cost of course scales with cells × domain diameter, so the saving is larger on bigger grids.

# Author's Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Fixed bug where the shading look-up table for MATLAB grids was pre-computed during initialization even when shading is disabled.  https://github.com/EBFMorg/EBFM/pull/155.
 * Fixed `phys["percolation"] = "uniform"` raising `TypeError` on the first timestep in `LOOP_SNOW.py` (only in NumPy path). `test_loop_snow_percolation.py` added.
 * EBFM now officially support and tests Python 3.14. https://github.com/EBFMorg/EBFM/pull/156
 * Fixed bug where resuming from a restart file (`--restart-init`) could lead to time step sizes equal to zero in `LOOP_SNOW.py`. Loading a restart file now asserts the restart variables to contain no missing values and converts them to plain `ndarray`. https://github.com/EBFMorg/EBFM/pull/154
--- a/src/ebfm/core/INIT.py
+++ b/src/ebfm/core/INIT.py
@@ -386,46 +386,48 @@ def init_grid(grid: GridDict, io, config: GridConfig):
         # -----------------------------------------------------------------------------------------------------
         # Pre-compute maximum grid elevation angle for various azimuth angles (needed for shading calculation)
         # -----------------------------------------------------------------------------------------------------
-        grid["shading_method"] = ShadingMethod.LUT  # shading based on look-up table (lut)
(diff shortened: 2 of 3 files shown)


src/ebfm/core/INIT.py
```diff
@@ -386,46 +386,48 @@ def init_grid(grid: GridDict, io, config: GridConfig):
         # -----------------------------------------------------------------------------------------------------
         # Pre-compute maximum grid elevation angle for various azimuth angles (needed for shading calculation)
         # -----------------------------------------------------------------------------------------------------
-        grid["shading_method"] = ShadingMethod.LUT  # shading based on look-up table (lut)
-        grid["nr_az_steps"] = 24  # number of azimuth angles (e.g. 24 = 1 per hour)
+        # Skip (expensive) pre-computation look-up table when shading is disabled.
+        if grid["has_shading"]:
```
BenjaminRodenberg: Introduce a small helper function `_precompute_shading_matlab` for everything inside the `if grid["has_shading"]`

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Fixed bug where the shading look-up table for MATLAB grids was pre-computed during initialization even when shading is disabled.  https://github.com/EBFMorg/EBFM/pull/155.
```
BenjaminRodenberg, line 9: Really a bug or performance optimization? I guess this depends a bit on the perspective ;)

If you classify this as bug: Please add the corresponding label in this PR.

A bugfix should ideally come with a test. But I assume testing this is not so easy; in that case we can also skip it.

BenjaminRodenberg (approved):
Two minor (optional) points.

src/ebfm/core/INIT.py
matthias-ibach: Good idea! Everything inside the `if` moved into `_precompute_shading_matlab(grid, mask_2D)`, next to the other shading helpers. The call site now only states the decision.

CHANGELOG.md
matthias-ibach: True, in this case I would call it an optimization :D

Building the look-up table during initialization (regardless of the fact that it was never read) lead to an initialization time of approx 25s for the `examples/dem_and_mask.mat` (instead of `0.19s` when turned off).

But still, I tried to come up with a test for the `--no-shading` option to check, since this can really become an issue for large grids. You can review the added test and comment if this fits here.

On the test: There are now two tests (taking about 1s), because neither test lets the table be actually built: The disabled case skips it by design, the enabled case patches the helper and only checks that it is called. What the table contains is untouched (this was not tested).. I verified both fail when the has_shading guard is removed.

PR #156, opened 2026-08-11
Add Python 3.14 to the tested versions

matthias-ibach:
Python 3.14 is the version provided by a miniforge on Levante and used by the TerraDT test case, but it was not covered by the CI here yet. This adds it to the `tox envlist`, the `gh-actions mapping` and the `unit-test` and `example-tests` matrices.

The MPI unit-test job stays on 3.13. It deliberately runs for a single version only and at the moment 3.13 is the most commonly used one. Also documents 3.11 as the version used on LUMI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Add Python 3.14 to the tested versions (unit and example tests). https://github.com/EBFMorg/EBFM/pull/156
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Add Python 3.14 to the tested versions (unit and example tests). https://github.com/EBFMorg/EBFM/pull/156
```
BenjaminRodenberg: This formulation is quite technical. Since the changelog is mainly intended for users I would "sell" it as a feature:

```suggestion
* EBFM now officially support and tests Python 3.14. https://github.com/EBFMorg/EBFM/pull/156
```

BenjaminRodenberg (approved):
Rest looks good to me.

We will probably remove support for Python 3.10 at some point in time (see #140); but this is not in scope for this PR and just a note.

PR #158, opened 2026-08-12
Fix percolation TypeError

matthias-ibach:
While adding coverage for the different physics mode to the GPU/NumPy comparison test (part of the GPU offloading MR) AI found the bug described below.

**Problem:**  

`phys["percolation"] = "uniform"` crashes on the first timestep of every run:

``` 
    carrot[np.arange(carrot.shape[0]), : ind + 1] = 1 / z0
    TypeError: only integer scalar arrays can be converted to a scalar index
``` 

The uniform scheme spreads the available water evenly over the layers down to the characteristic percolation depth  `ind` is the index of that layer per column, i.e. an array of length `gpsum`, this cannot be the stop of a slice. NumPy gets `slice(None, array([...]))` and rejects it. It is a variable-length range per row, which slicing cannot express.
Only the NumPy path is affected. The Numba kernel already selects the layers correctly with `k <= ind`.

**Fix:**

```
carrot[np.arange(carrot.shape[1])[None, :] <= ind[:, None]] = 1 / z0
```

Select the layers with a mask, matching what the kernel does. Verified with one `LOOP_SNOW` step on a 48x50 grid (artificial grid created with AI assistance). The mode now runs and the NumPy path agrees with the Numba kernel on every field.

**Background (found while digging and consulting the AI):**

The line is unchanged since the initial commit `42c0113` (2025-09-04; commit `31a8311` only reformatted it). It reads like a MATLAB translation artefact: there the loop is per gridpoint, `ind` is a scalar and `1:ind+1` is valid. Vectorizing over columns broke it. It survived because nothing ever selects the mode: `phys["percolation"]` is hardcoded to "normal" in `INIT.py` and is not exposed on the CLI, so reaching "uniform" requires editing the source. No test, example run or benchmark ever
did.

# Author's Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Fixed `phys["percolation"] = "uniform"` raising `TypeError` on the first timestep in `LOOP_SNOW.py` (only in NumPy path). `test_loop_snow_percolation.py` added.
 * EBFM now officially support and tests Python 3.14. https://github.com/EBFMorg/EBFM/pull/156
-* Fixed `phys["percolation"] = "uniform"` raising `TypeError` on the first timestep in `LOOP_SNOW.py` (only in NumPy path).
 * Fixed bug where resuming from a restart file (`--restart-init`) could lead to time step sizes equal to zero in `LOOP_SNOW.py`. Loading a restart file now asserts the restart variables to contain no missing values and converts them to plain `ndarray`. https://github.com/EBFMorg/EBFM/pull/154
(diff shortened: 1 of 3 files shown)


BenjaminRodenberg (changes requested):
Hard to judge for me if this is actually a bug and what the fix changes.

We definitely need a test here (AI should be able to do this quite quickly). As far as I understand it this bug has never been noticed because the code is usually not executed. I assume this should also help us to get a better idea what was wrong originally and what this fix does.

Please don't forget to also ensure that the test fails if the fix is not present; I do this by just locally reverting the commit with the fix and running the test.

BenjaminRodenberg (approved):
* I checked that the test is sensitive (if I revert the fix it fails)
* Content-wise I only briefly looked over the test; but since we will remove the feature soon that's good enough in my opinion.

PR #160, opened 2026-08-12
Remove "uniform" percolation scheme

### LINKED ISSUE #159: Remove "uniform" percolation scheme
`phys["percolation"]` offers four schemes (see [Marchenko et al. 2017](https://www.frontiersin.org/journals/earth-science/articles/10.3389/feart.2017.00016/full)): `bucket`,`normal`, `linear`, `uniform`. In practice only `normal` and `bucket` are used. `uniform` has never been tested or used in the Python code. @wardvp  proposed to remove it since it is not really used. 

It was also broken before: the NumPy implementation raised `TypeError` on the first timestep and has done that since the initial commit (42c0113). `phys["percolation"]` is hardcoded in `INIT.py` and not reachable from the command line, so the branch has never been executed. It surfaced while adding per-mode coverage to the GPU/NumPy comparison test (see #157).

A fix with a proper test was introduced in #158. The changes will have to be reverted/adapted once this feature is removed.

matthias-ibach:
Closes #159. 

`uniform` is one of four percolation schemes in `phys["percolation"]`. Discussions in #158 revealed the fact that this scheme is never used, nor has it ever been tested in the Python code. So the decision was to remove this feature from the code.

It was also broken from the initial commit until #158 (the NumPy path raised `TypeError` on the first timestep). This went unnoticed because `phys["percolation"]` is hardcoded in `INIT.py` and therefore not reachable from the CLI, so
the branch never ran. 

**What changed:**
- The scheme is dropped from the NumPy path, from the Numba kernel (mode 3) and from the option list in `INIT.py`.
- Selecting `uniform` now raises the same `ValueError` as any other unknown scheme, rather than falling through
- `tests/core/test_loop_snow_percolation.py`, added in #158  as a smoke test for the scheme, now pins which schemes are accepted instead

# Author's Todos

### DIFF
--- a/src/ebfm/core/LOOP_SNOW_kernels.py
+++ b/src/ebfm/core/LOOP_SNOW_kernels.py
@@ -330,6 +330,10 @@ def _percolation_kernel(
     6. Irreducible-water refreezing
     """
     gpsum, nl = subT.shape
+    # Checked here and not in the carrot branch below: a raise inside the prange body is a second loop exit,
+    # which makes Numba run the kernel serially.
+    if percolation_mode < 0 or percolation_mode > 2:
+        raise RuntimeError("_percolation_kernel: unknown percolation_mode, expected 0=bucket, 1=normal, 2=linear")
     sigma2_2 = 2.0 * (perc_depth / 3.0) ** 2
     norm_coeff = 2.0 / (perc_depth / 3.0) / math.sqrt(2.0 * math.pi)
     trunoff_factor = 1.0 / (1.0 + dt / Trunoff)
@@ -370,7 +374,7 @@ def _percolation_kernel(
                 zz_k = depth + 0.5 * subZ[i, k]
                 if percolation_mode == 1:  # normal (Gaussian)
                     carrot_loc[k] = norm_coeff * math.exp(-(zz_k * zz_k) / sigma2_2)
-                else:  # linear (mode 2)
+                elif percolation_mode == 2:  # linear
                     v = 2.0 * (perc_depth - zz_k) / (perc_depth * perc_depth)
                     carrot_loc[k] = v if v > 0.0 else 0.0
                 depth += subZ[i, k]
--- a/src/ebfm/core/LOOP_SNOW_kernels.py
+++ b/src/ebfm/core/LOOP_SNOW_kernels.py
@@ -242,13 +242,13 @@ def _percolation_kernel(
     Lm,
     Trunoff,
     perc_depth,
-    percolation_mode,  # 0=bucket, 1=normal, 2=linear, 3=uniform
+    percolation_mode,  # 0=bucket, 1=normal, 2=linear
     dt,
 ):
     """Per-column percolation, slush storage and refreezing kernel, parallelized over gpsum
 
     1. Compute Wlim and Wirr (refreezing potential and available irreducible water storate)
-    2. carrot distribution profile (bucket / normal / linear / uniform)
+    2. carrot distribution profile (bucket / normal / linear)
     3. Refreezing and irreducible-water-storage
     4. Slush storage
     5. Slush refreezing
@@ -285,29 +285,16 @@ def _percolation_kernel(
         if percolation_mode == 0:  # bucket: all water enters surface layer
             carrot_loc[0] = 1.0
         else:
-            # Compute zz (midpoint depth of each layer) for mode 1/2/3
+            # Compute zz (midpoint depth of each layer) for mode 1/2
             depth = 0.0
             for k in range(nl):
                 zz_k = depth + 0.5 * subZ[i, k]
                 if percolation_mode == 1:  # normal (Gaussian)
                     carrot_loc[k] = norm_coeff * math.exp(-(zz_k * zz_k) / sigma2_2)
-                elif percolation_mode == 2:  # linear
+                else:  # linear (mode 2)
                     v = 2.0 * (perc_depth - zz_k) / (perc_depth * perc_depth)
                     carrot_loc[k] = v if v > 0.0 else 0.0
-                else:  # uniform (mode 3): temporarily store zz for the argmin pass below
-                    carrot_loc[k] = zz_k
                 depth += subZ[i, k]
-
-            if percolation_mode == 3:  # uniform: resolve argmin, then fill layers 0..ind
-                min_dist = math.inf
-                ind = 0
-                for k in range(nl):
-                    d = abs(carrot_loc[k] - perc_depth)
-                    if d < min_dist:
-                        min_dist = d
-                        ind = k
-                for k in range(nl):
-                    carrot_loc[k] = (1.0 / perc_depth) if k <= ind else 0.0
 
         # Scale by layer thickness, normalize, multiply by avail_W
         s = 0.0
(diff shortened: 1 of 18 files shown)


src/ebfm/core/LOOP_SNOW_kernels.py
```diff
@@ -285,30 +285,17 @@ def _percolation_kernel(
...
         else:
-            # Compute zz (midpoint depth of each layer) for mode 1/2/3
+            # Compute zz (midpoint depth of each layer) for mode 1/2
             depth = 0.0
             for k in range(nl):
                 zz_k = depth + 0.5 * subZ[i, k]
                 if percolation_mode == 1:  # normal (Gaussian)
                     carrot_loc[k] = norm_coeff * math.exp(-(zz_k * zz_k) / sigma2_2)
-                elif percolation_mode == 2:  # linear
+                else:  # linear (mode 2)
```
BenjaminRodenberg, line 373: keep `elif perlocation_mode == 2` and `raise RuntimeException(...)` on `else:`. That's safer IMO

tests/core/test_loop_snow_percolation.py
```diff
@@ -1,33 +1,20 @@
...
-characteristic percolation depth C["perc_depth"] and adds nothing below it.
-That cut-off layer is determined per column, so columns with different layer
-thicknesses wet down to different layer indices.
+The percolation schemes LOOP_SNOW accepts.
 
-The default configuration uses percolation="normal" and the scheme is not
-reachable from the command line, so this code path had never been executed.
+"bucket", "normal" and "linear" run. Anything else (including the removed
+"uniform" scheme) is rejected with a clear error instead of being silently
+ignored.
```
BenjaminRodenberg, line 10:
```suggestion
"bucket", "normal" and "linear" run.
```
No need to talk about the history here. If somebody is really interested there is git.

tests/core/test_loop_snow_percolation.py
```diff
@@ -42,7 +29,7 @@ def _make_case(C):
         "mask": np.ones(GPSUM, dtype=int),
     }
     OUT = {
-        "subZ": np.repeat(_THICKNESS[:, None], NL, axis=1),
+        "subZ": np.full((GPSUM, NL), 1.0),
```
BenjaminRodenberg, line 32: unrelated?

tests/core/test_loop_snow_percolation.py
```diff
@@ -71,50 +58,31 @@ def _make_case(C):
...
-    """percolation="uniform" must run and must stop at the percolation depth."""
-
+class TestPercolationSchemes(unittest.TestCase):
     def setUp(self):
         from ebfm.core import INIT, LOOP_SNOW
 
         self.LOOP_SNOW = LOOP_SNOW
         self.C = INIT.init_constants()
-        self.phys = {"snow_compaction": "firn+snow", "percolation": "uniform"}
-        self.grid, self.OUT, self.IN = _make_case(self.C)
```
BenjaminRodenberg, line 83: Not needed anymore?

tests/core/test_loop_snow_percolation.py
```diff
@@ -71,50 +58,31 @@ def _make_case(C):
...
+            with self.subTest(percolation=percolation):
+                OUT = self._run(percolation)
+                self.assertTrue(np.isfinite(OUT["subT"]).all())
+
+    def test_removed_and_unknown_schemes_are_rejected(self):
+        """"uniform" was removed; selecting it must fail, not fall through."""
+        for percolation in ("uniform", "not-a-scheme"):
+            with self.subTest(percolation=percolation):
+                with self.assertRaises(ValueError):
+                    self._run(percolation)
```
BenjaminRodenberg, line 85: Not really needed in my opinion. Sufficient if this is checked manually once during development.

tests/core/test_loop_snow_percolation.py
```diff
@@ -71,50 +58,31 @@ def _make_case(C):
...
-        for column, cutoff in enumerate(_EXPECTED_CUTOFF):
-            with self.subTest(column=column, thickness=_THICKNESS[column]):
-                self.assertTrue(
-                    warmed[column, : cutoff + 1].all(),
-                    f"column {column}: layers 0..{cutoff} should have received water, got {warmed[column]}",
-                )
-                self.assertFalse(
-                    warmed[column, cutoff + 1 :].any(),
-                    f"column {column}: no layer below {cutoff} should have received water, got {warmed[column]}",
-                )
```
BenjaminRodenberg, line 114: Possible to adapt the tests to one of the schemes we still support?

What we would sacrifice if you just delete these tests: `self.LOOP_SNOW.main` is also shared by all perlocation schemes and would be uncovered if you remove the test. `test_uniform_percolation_runs`, for example, is really a minimal test and should also work for the other three schemes I assume (this would actually even improve our coverage).

Maybe something the AI can do if you provide this test?

BenjaminRodenberg (changes requested):
If it is possible to recycle the tests for the perlocation schemes we still support that would be nice.

Removal of the uniform perlocation is generally fine.

matthias-ibach (commented):
> If it is possible to recycle the tests for the perlocation schemes we still support that would be nice.
> 
> Removal of the uniform perlocation is generally fine.

Good idea! I recycled rather than deleted the perlocation test. `test_supported_schemes_run` and `test_rain_warms_the_column` now run all three schemes through `LOOP_SNOW.main` and the old uniform depth test lives on as `test_linear_percolation_stops_at_percolation_depth` (this was done with the help of AI, see comment at the end).

BenjaminRodenberg (approved):
Looks good to me from the technical side.

Nice side effect that we now have tests for the remaining schemes.

Just to clarify: We still wait for feedback from @wardvp before merging this.

matthias-ibach:
> Looks good to me from the technical side.
> 
> Nice side effect that we now have tests for the remaining schemes.
> 
> Just to clarify: We still wait for feedback from @wardvp before merging this.

Thanks for the review!
And yes, I would wait until we get the go before merging this.

src/ebfm/core/LOOP_SNOW_kernels.py
matthias-ibach: Yes, and that was not basic knowledge for me either 😄  
I actually measured it (using AI assistance): the fallback is decided at compile time, so even a raise that never fires makes the loop run single-threaded (it was about 7x slower in a quick 8-thread benchmark). Same for `assert` or break in a `prange` body. However, exits from inner loops are actually fine, so the break in `_heat_conduction_kernel` is unaffected. That is why added a a comment there (so it will not be re-added later).

wardvp (approved):
This all looks good to me! Thanks for the efforts to remove this feature. I will do some tests later on to see if the other percolation schemes are working as expected.

BenjaminRodenberg:
> This all looks good to me! Thanks for the efforts to remove this feature. I will do some tests later on to see if the other percolation schemes are working as expected.

@wardvp thanks for the feedback and checking! We would then wait for your tests before merging. Can you drop us a quick note when you are done?

wardvp:
> > This all looks good to me! Thanks for the efforts to remove this feature. I will do some tests later on to see if the other percolation schemes are working as expected.
> 
> @wardvp thanks for the feedback and checking! We would then wait for your tests before merging. Can you drop us a quick note when you are done?

I checked the other schemes and they seem to function normally.

matthias-ibach:
> > > This all looks good to me! Thanks for the efforts to remove this feature. I will do some tests later on to see if the other percolation schemes are working as expected.
> > 
> > 
> > @wardvp thanks for the feedback and checking! We would then wait for your tests before merging. Can you drop us a quick note when you are done?
> 
> I checked the other schemes and they seem to function normally.

Great, thanks a lot for checking this @wardvp!  
I will then go ahead, resolve the conflicts and merge this branch into main.

PR #161, opened 2026-08-18
Improve performance documentation

matthias-ibach:
Follow-up to #157, documentation only, no code changes.

The `README` had grown a large performance section: Numba threading, GPU offloading, timing, snapshot comparison, reproducibility flags, plus a GPU-specific troubleshooting entry. All of that was moved into `docs/Performance.md` so the README stays an overview.

# Author's Todos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
-* Moved performance and profiling sections out of `README.md` into `docs/Performance.md`, together with the GPU setup and CUDA runtime troubleshooting entry. https://github.com/EBFMorg/EBFM/pull/XXX
+* Moved performance and profiling sections out of `README.md` into `docs/Performance.md`, together with the GPU setup and CUDA runtime troubleshooting entry. https://github.com/EBFMorg/EBFM/pull/161
 * Removed the `uniform` percolation scheme (not used in practice, `normal` and `bucket` are typically chosen). `phys["percolation"]` now accepts `bucket`, `normal` and `linear`. Adapted tests. https://github.com/EBFMorg/EBFM/pull/160
 * Fixed `phys["percolation"] = "uniform"` raising `TypeError` on the first timestep in `LOOP_SNOW.py` (only in NumPy path). `test_loop_snow_percolation.py` added. https://github.com/EBFMorg/EBFM/pull/158
 * Introduced `--with-gpu` to offload the kernels compaction, heat conduction, percolation to a GPU via `numba.cuda` (NVIDIA) or `numba.hip` (AMD), with `--gpu-vendor {auto,nvidia,amd}` to select/guard the vendor. Mutually exclusive with `--with-numba`. Install via `pip install ebfm[gpu]`. https://github.com/EBFMorg/EBFM/pull/157
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Moved performance and profiling sections out of `README.md` into `docs/Performance.md`, together with the GPU setup and CUDA runtime troubleshooting entry. https://github.com/EBFMorg/EBFM/pull/XXX
 * Removed the `uniform` percolation scheme (not used in practice, `normal` and `bucket` are typically chosen). `phys["percolation"]` now accepts `bucket`, `normal` and `linear`. Adapted tests. https://github.com/EBFMorg/EBFM/pull/160
 * Fixed `phys["percolation"] = "uniform"` raising `TypeError` on the first timestep in `LOOP_SNOW.py` (only in NumPy path). `test_loop_snow_percolation.py` added. https://github.com/EBFMorg/EBFM/pull/158
 * Introduced `--with-gpu` to offload the kernels compaction, heat conduction, percolation to a GPU via `numba.cuda` (NVIDIA) or `numba.hip` (AMD), with `--gpu-vendor {auto,nvidia,amd}` to select/guard the vendor. Mutually exclusive with `--with-numba`. Install via `pip install ebfm[gpu]`. https://github.com/EBFMorg/EBFM/pull/157
--- a/docs/Performance.md
+++ b/docs/Performance.md
@@ -1,0 +1,121 @@
+<!--
+SPDX-FileCopyrightText: 2026 EBFM Authors
+
+SPDX-License-Identifier: BSD-3-Clause
+-->
+
+# Performance and Profiling Runs
+
+## Installation for Performance Runs
+
+Install EBFM with the performance features:
+
+```sh
+pip install -e .[performance]
+```
+
+This additionally installs `numba` to run with multiple CPU-threads.
+
+## Running EBFM with performance optimizations
+
+EBFM now includes several performance improvements and supports several options for performance testing and benchmarking:
+
+- Regular (NumPy) path:
+
+  The default run already includes some performance improvements. Run as usual:
+
+  ```sh
+  ebfm --matlab-mesh examples/dem_and_mask.mat
+  ```
+
+- Numba kernels:
+
+  To enable Numba-accelerated kernels, use the `--with-numba` flag. You can control the number of threads with `--numba-threads N` (replace `N` with the desired thread count):
+
+  ```sh
+  ebfm --matlab-mesh examples/dem_and_mask.mat --with-numba --numba-threads 2
+  ```
+
+  Note: If you use more than one thread, you must specify `--numba-threads`. In practice, 2 threads have shown the best performance so far, but optimal settings depend on your hardware and problem size. Feel free to experiment.
+
+- GPU offloading:
+
+  Install the GPU dependencies and use the `--with-gpu` flag to offload kernels from `LOOP_SNOW.py` to a GPU:
+
+  ```sh
+  pip install -e .[gpu]
+  ebfm --matlab-mesh examples/dem_and_mask.mat --with-gpu
+  ```
+
+  The same kernels run on NVIDIA (via `numba.cuda`) and on AMD (via `numba.hip`). The vendor is detected automatically. You can use `--gpu-vendor {auto,nvidia,amd}` to select or guard it explicitly. `--with-gpu` and `--with-numba` are mutually exclusive.
+
+  Note: on a cluster you usually have to load a CUDA/ROCm toolkit and point `CUDA_HOME` at it, otherwise Numba cannot compile the kernels even though the GPU itself is visible. See [CUDA runtime not found](#numba-does-not-find-the-cuda-runtime-cudais_available-is-false) in the troubleshooting section.
+
+  Before a long run you can verify the setup with:
+
+  ```sh
+  python -c "from numba import cuda; print(cuda.is_available())"
+  ```
+
+  At startup `--with-gpu` reports which stack and device it found, so you can check it went to the GPU you expected:
+
+  ```
+  [GPU] backend enabled (nvidia). Device: NVIDIA A100-SXM4-80GB  free=79.15 GiB  total=79.15 GiB
+  ```
+
+## Timing Your Run
+
+To measure the total runtime, simply prepend your command with `time`:
+
+```sh
+time ebfm --matlab-mesh examples/dem_and_mask.mat --with-numba --numba-threads 2
+```
+
+## Comparing Results
+
+A script is provided to compare model output snapshots:
+
+- `tools/compare_snapshots.py` can be used to compare two output files (e.g., from different runs or configurations).
+
+To create a reference file for comparison, use the `--dump-reference` option at the end of your run:
+
+```sh
+ebfm --matlab-mesh examples/dem_and_mask.mat --dump-reference reference_run.npz
+```
+
+Then compare with:
+
+```sh
+python tools/compare_snapshots.py reference_run.npz new_run.npz
+```
+
+Note: If you use the random-forcing within EBFM in your testcases, make sure to additionally set the option `--random-seed`, as explained below.
+
+## Diagnostics and Reproducibility
+
+- The option `--random-seed` allows reproducible runs (especially if the random forcing in the example testcase is used). Set a fixed seed to ensure identical results for repeated runs (important for benchmarking and debugging):
+
+  ```sh
+  ebfm --matlab-mesh examples/dem_and_mask.mat --random-seed 42
+  ```
+
+- Use `--diagnostics` to print some diagnostics for a quick overview for every timestep:
+
+  ```sh
+  ebfm --matlab-mesh examples/dem_and_mask.mat --diagnostics
+  ```
+
+## Troubleshooting
+
+### Numba does not find the CUDA runtime (`cuda.is_available()` is `False`)
+
+*Problem:* `--with-gpu` fails, or `python -c "from numba import cuda; print(cuda.is_available())"` prints `False`, even though the GPU is visible. Note that `cuda.detect()` and `nvidia-smi` can still report the device correctly: they only need the CUDA *driver*, while Numba additionally needs the CUDA *runtime* (`libnvvm`, `libcudart`) to compile kernels.
+
+*Solution:* Point `CUDA_HOME` at a CUDA toolkit installation and add its libraries to `LD_LIBRARY_PATH`. On Levante, for example:
+
+```sh
+export CUDA_HOME=/sw/spack-levante/nvhpc-22.5-v4oky3/Linux_x86_64/22.5/cuda/11.7
+export LD_LIBRARY_PATH=$CUDA_HOME/lib64:$LD_LIBRARY_PATH
+```
+
+Afterwards `cuda.is_available()` should print `True`.
--- a/docs/Performance.md
+++ b/docs/Performance.md
@@ -11,47 +11,16 @@ SPDX-License-Identifier: BSD-3-Clause
 Install EBFM with the performance features:
 
 ```sh
-pip install -e .[performance]
+pip install ebfm[performance]
 ```
 
 This additionally installs `numba` to run with multiple CPU-threads.
 
-## Running EBFM with performance optimizations
-
-EBFM offers several options for tuning performance on a given system and for testing and benchmarking it.
-
-### Numba kernels: `--with-numba`
-
-`--with-numba` runs the `LOOP_SNOW.py` kernels (compaction, heat conduction, percolation) as compiled,
-multi-threaded Numba kernels instead of NumPy array operations. Use it on a multi-core CPU when no GPU is available.
-It generally outperforms the NumPy path, the more clearly the larger the mesh and the longer the run.
-
-It is not on by default because `numba` is an optional dependency and the kernels compile on first use. A very short run can therefore spend more time compiling than it saves.
-
-Control the number of threads with `--numba-threads N` (replace `N` with the desired thread count):
+For GPU offloading, install the GPU dependencies instead:
 
 ```sh
-ebfm --matlab-mesh examples/dem_and_mask.mat --with-numba --numba-threads 2
+pip install ebfm[gpu]
 ```
-
-Note: If you use more than one thread, you must specify `--numba-threads`. In practice, 2 threads have shown the
-best performance so far, but optimal settings depend on your hardware and problem size. Feel free to experiment.
-
-### GPU offloading: `--with-gpu`
-
-You can use this option when the node you run on has a GPU: it offloads the same kernels to the device. How much that
-gains depends on the mesh size and the GPU, and for small meshes it is not necessarily faster than `--with-numba`.
-`--with-gpu` and `--with-numba` are alternative backends for those kernels and therefore mutually exclusive.
-
-Install the GPU dependencies and run with the `--with-gpu` flag:
-
-```sh
-pip install -e .[gpu]
-ebfm --matlab-mesh examples/dem_and_mask.mat --with-gpu
-```
-
-The same kernels run on NVIDIA (via `numba.cuda`) and on AMD (via `numba.hip`). The vendor is detected
-automatically. You can use `--gpu-vendor {auto,nvidia,amd}` to select or guard it explicitly.
 
 Note: on a cluster you usually have to load a CUDA/ROCm toolkit and point `CUDA_HOME` at it, otherwise Numba
 cannot compile the kernels even though the GPU itself is visible. See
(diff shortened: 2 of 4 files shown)


docs/Performance.md
```diff
@@ -0,0 +1,121 @@
...
+
+```sh
+pip install -e .[performance]
+```
+
+This additionally installs `numba` to run with multiple CPU-threads.
+
+## Running EBFM with performance optimizations
+
+EBFM now includes several performance improvements and supports several options for performance testing and benchmarking:
```
BenjaminRodenberg:
```suggestion
EBFM offers several options for tuning its performance on specific systems and it supports several options for performance testing and benchmarking:
```
I'd rather write this in the style of general docs than in the style of a report (actual progress is not so relevant here but rather which features exist and how to use them)

docs/Performance.md
```diff
@@ -0,0 +1,121 @@
...
+
+- Regular (NumPy) path:
+
+  The default run already includes some performance improvements. Run as usual:
+
+  ```sh
+  ebfm --matlab-mesh examples/dem_and_mask.mat
+  ```
```
BenjaminRodenberg:
```suggestion
```
Nice to have this feature but the user does not really need to know because it comes out-of-the-box

docs/Performance.md
```diff
@@ -0,0 +1,121 @@
...
+
+- Numba kernels:
+
+  To enable Numba-accelerated kernels, use the `--with-numba` flag. You can control the number of threads with `--numba-threads N` (replace `N` with the desired thread count):
+
+  ```sh
+  ebfm --matlab-mesh examples/dem_and_mask.mat --with-numba --numba-threads 2
+  ```
+
+  Note: If you use more than one thread, you must specify `--numba-threads`. In practice, 2 threads have shown the best performance so far, but optimal settings depend on your hardware and problem size. Feel free to experiment.
```
BenjaminRodenberg: Maybe help the user here a bit: When does it make sense to activate this option? Why is it not turned on by default?

docs/Performance.md
```diff
@@ -0,0 +1,121 @@
...
+
+- GPU offloading:
```
BenjaminRodenberg: Motivation is probably more obvious: If you are running on a GPU; use this option. But still useful to explicitly write it here in a sentence.

docs/Performance.md
```diff
@@ -0,0 +1,121 @@
...
+  Install the GPU dependencies and use the `--with-gpu` flag to offload kernels from `LOOP_SNOW.py` to a GPU:
+
+  ```sh
+  pip install -e .[gpu]
+  ebfm --matlab-mesh examples/dem_and_mask.mat --with-gpu
+  ```
+
+  The same kernels run on NVIDIA (via `numba.cuda`) and on AMD (via `numba.hip`). The vendor is detected automatically. You can use `--gpu-vendor {auto,nvidia,amd}` to select or guard it explicitly. `--with-gpu` and `--with-numba` are mutually exclusive.
+
+  Note: on a cluster you usually have to load a CUDA/ROCm toolkit and point `CUDA_HOME` at it, otherwise Numba cannot compile the kernels even though the GPU itself is visible. See [CUDA runtime not found](#numba-does-not-find-the-cuda-runtime-cudais_available-is-false) in the troubleshooting section.
```
BenjaminRodenberg: Move this troubleshooting bit from the root level README to this one?

Drawback:
- Troubleshooting will be scattered throughout the docs

Advantage:
- The lengthy root level README becomes shorter
- Topic-wise it belongs here

docs/Performance.md
```diff
@@ -0,0 +1,121 @@
...
+
+  Before a long run you can verify the setup with:
```
BenjaminRodenberg:
```suggestion
  It is recommended to verify your cuda setup by running:
```

docs/Performance.md
```diff
@@ -0,0 +1,121 @@
...
+
+  ```sh
+  python -c "from numba import cuda; print(cuda.is_available())"
+  ```
```
BenjaminRodenberg:
````suggestion
  ```sh
  python -c "from numba import cuda; print(cuda.is_available())"
  # expected output: ...
  ```
````

I would add the expected output here; this allows the user to check if it's working as expected. This is done very nicely in the NumPy docs, for example:

```sh
python3
Python 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> help(numpy.array)
...
    Examples
    --------
    >>> np.array([1, 2, 3])
    array([1, 2, 3])

    Upcasting:

    >>> np.array([1, 2, 3.0])
    array([ 1.,  2.,  3.])
```

docs/Performance.md
```diff
@@ -0,0 +1,121 @@
...
+
+  At startup `--with-gpu` reports which stack and device it found, so you can check it went to the GPU you expected:
+
+  ```
+  [GPU] backend enabled (nvidia). Device: NVIDIA A100-SXM4-80GB  free=79.15 GiB  total=79.15 GiB
+  ```
+
+## Timing Your Run
+
+To measure the total runtime, simply prepend your command with `time`:
```
BenjaminRodenberg:
```suggestion
To measure the total runtime, simply prepend your command with [`time`](TODO):
```
Add a link to the docs of this tool? Useful if things change in the future.

docs/Performance.md
```diff
@@ -0,0 +1,121 @@
...
+```sh
+time ebfm --matlab-mesh examples/dem_and_mask.mat --with-numba --numba-threads 2
+```
+
+## Comparing Results
+
+A script is provided to compare model output snapshots:
+
+- `tools/compare_snapshots.py` can be used to compare two output files (e.g., from different runs or configurations).
+
```
BenjaminRodenberg:
```suggestion
The script `tools/compare_snapshots.py` allows to compare two output files (e.g., from different runs or configurations).

```

docs/Performance.md
```diff
@@ -0,0 +1,121 @@
...
+*Problem:* `--with-gpu` fails, or `python -c "from numba import cuda; print(cuda.is_available())"` prints `False`, even though the GPU is visible. Note that `cuda.detect()` and `nvidia-smi` can still report the device correctly: they only need the CUDA *driver*, while Numba additionally needs the CUDA *runtime* (`libnvvm`, `libcudart`) to compile kernels.
+
+*Solution:* Point `CUDA_HOME` at a CUDA toolkit installation and add its libraries to `LD_LIBRARY_PATH`. On Levante, for example:
+
+```sh
+export CUDA_HOME=/sw/spack-levante/nvhpc-22.5-v4oky3/Linux_x86_64/22.5/cuda/11.7
+export LD_LIBRARY_PATH=$CUDA_HOME/lib64:$LD_LIBRARY_PATH
+```
+
+Afterwards `cuda.is_available()` should print `True`.
```
BenjaminRodenberg: Provide a source code snipped instead?

```sh
python -c "import cuda; print(cude.is_available())
# expected output: True
```

CHANGELOG.md
```diff
@@ -6,8 +6,12 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # develop
 
+* Moved performance and profiling sections out of `README.md` into `docs/Performance.md`, together with the GPU setup and CUDA runtime troubleshooting entry.
+https://github.com/EBFMorg/EBFM/pull/XXX
```
BenjaminRodenberg:
```suggestion
https://github.com/EBFMorg/EBFM/pull/161
```

docs/Performance.md
matthias-ibach: Yes, agreed. I have rewritten it in a style of a general documentation rather than that of a report. Feel free to check if you like this better.

docs/Performance.md
matthias-ibach: Alright, I removed it.

docs/Performance.md
matthias-ibach: Good pojnt, I extended the documentation here.  
It now says to use it on a multi-core CPU when no GPU is available and states why it is not set per default: numba is an optional dependency and the kernels compile on first use, so a very short run can spend more compiling than it saves.

docs/Performance.md
matthias-ibach: Yes, I tried to make it more explicit here.

docs/Performance.md
matthias-ibach: I agree with you. My thinking was actually the same as yours.  
I decided to go ahead and add a reference in the `README` so it is still discoverable, but the actual solution in the `doc/Performance.md`, where it's greppable in context and only shows up for people using the feature.  
If you think that arranging it differently is better, feel free to comment!

docs/Performance.md
matthias-ibach: Thanks, changed.

docs/Performance.md
matthias-ibach: Good point.   
I added a link to the Bash manual. 

Additionally, I added a block to show what output the user can expect.

docs/Performance.md
matthias-ibach: Done, I added the expected output both here and at the first check earlier in the page.

docs/Performance.md
matthias-ibach: Done, thank you for the suggestion.

docs/Performance.md
matthias-ibach: Done, see comment above.

CHANGELOG.md
matthias-ibach: Changed.

docs/Performance.md
BenjaminRodenberg: Sounds good. I'll close this discussion.

Comment on the general design of the interface: I remember that FEniCS also uses some form of JIT by default; this consumes quite some time on the first run but is turned on by default. So if the only reason for keeping it off by default is the one time cost one could also consider keeping it on by default to simplify the user interface.

But I have no idea what "very short run" vs. compile time means here. So I cannot really judge. If you think this would make sense: Feel free to open an issue; I can take care of changing the defaults and evaluating this further.

docs/Performance.md
BenjaminRodenberg: I don't have a strong opinion here. So let's keep things as they are. That's less work to do ;)

docs/Performance.md
```diff
@@ -0,0 +1,142 @@
...
+
+This additionally installs `numba` to run with multiple CPU-threads.
+
+## Running EBFM with performance optimizations
+
+EBFM offers several options for tuning performance on a given system and for testing and benchmarking it.
+
+### Numba kernels: `--with-numba`
+
+`--with-numba` runs the `LOOP_SNOW.py` kernels (compaction, heat conduction, percolation) as compiled,
```
BenjaminRodenberg:
```suggestion
`ebfm --with-numba` runs the `LOOP_SNOW.py` kernels (compaction, heat conduction, percolation) as compiled,
```

or

```suggestion
The option `--with-numba` runs the `LOOP_SNOW.py` kernels (compaction, heat conduction, percolation) as compiled,
```

Not really needed; but I found it a bit odd on first sight how the sentence starts. Feel free to ignore if you do not think this revision improves it.

docs/Performance.md
matthias-ibach: Sounds good, then we will leave it as it is for the moment.  
The good thing is: we can always change it :smile:

docs/Performance.md
```diff
@@ -0,0 +1,142 @@
...
+```sh
+ebfm --matlab-mesh examples/dem_and_mask.mat --with-numba --numba-threads 2
+```
+
+Note: If you use more than one thread, you must specify `--numba-threads`. In practice, 2 threads have shown the
+best performance so far, but optimal settings depend on your hardware and problem size. Feel free to experiment.
+
+### GPU offloading: `--with-gpu`
+
+You can use this option when the node you run on has a GPU: it offloads the same kernels to the device. How much that
```
BenjaminRodenberg:
```suggestion
You can use the option `--with-gpu` when the node you run on has a GPU: it offloads the same kernels to the device. How much that
```
I would repeat the option here to make it more clear.

docs/Performance.md
```diff
@@ -0,0 +1,142 @@
...
+gains depends on the mesh size and the GPU, and for small meshes it is not necessarily faster than `--with-numba`.
+`--with-gpu` and `--with-numba` are alternative backends for those kernels and therefore mutually exclusive.
```
BenjaminRodenberg:
```suggestion
The options `--with-gpu` and `--with-numba` tell EBFM to select alternative backends for those kernels and are therefore mutually exclusive.
```

docs/Performance.md
BenjaminRodenberg:
```suggestion
You can use this option when the node you run on has a GPU: it offloads the same kernels to the device as for the option `--with-numba`. How much that
```

docs/Performance.md
```diff
@@ -0,0 +1,142 @@
...
+multi-threaded Numba kernels instead of NumPy array operations. Use it on a multi-core CPU when no GPU is available.
+It generally outperforms the NumPy path, the more clearly the larger the mesh and the longer the run.
+
+It is not on by default because `numba` is an optional dependency and the kernels compile on first use. A very short run can therefore spend more time compiling than it saves.
+
+Control the number of threads with `--numba-threads N` (replace `N` with the desired thread count):
+
+```sh
+ebfm --matlab-mesh examples/dem_and_mask.mat --with-numba --numba-threads 2
+```
```
BenjaminRodenberg:
```suggestion
**Example:** Run EBFM with xyz:
```sh
ebfm --matlab-mesh examples/dem_and_mask.mat --with-numba --numba-threads 2
```
```

File: docs/Performance.md [hunk 16]
```diff
@@ -0,0 +1,142 @@
...
+You can use this option when the node you run on has a GPU: it offloads the same kernels to the device. How much that
+gains depends on the mesh size and the GPU, and for small meshes it is not necessarily faster than `--with-numba`.
+`--with-gpu` and `--with-numba` are alternative backends for those kernels and therefore mutually exclusive.
+
+Install the GPU dependencies and run with the `--with-gpu` flag:
+
+```sh
+pip install -e .[gpu]
+ebfm --matlab-mesh examples/dem_and_mask.mat --with-gpu
+```
```
BenjaminRodenberg:
````suggestion
**Example**: Run EBFM with GPU xyz

Note: You have to install the GPU dependencies first. Then run using the `--with-gpu` flag:

```sh
pip install -e .[gpu]
ebfm --matlab-mesh examples/dem_and_mask.mat --with-gpu
```
````

If you want: You can also move the installation step into the introduction of this paragraph since it only has to be done once (similar to `pip install -e .[performance]` above). Looks a bit out of place here.

File: docs/Performance.md [hunk 17]
```diff
@@ -0,0 +1,142 @@
...
+
+The same kernels run on NVIDIA (via `numba.cuda`) and on AMD (via `numba.hip`). The vendor is detected
+automatically. You can use `--gpu-vendor {auto,nvidia,amd}` to select or guard it explicitly.
```
BenjaminRodenberg:
```suggestion
```
Needed? This looks to me like something where the `--help` should be sufficient and no additional docs are needed.

docs/Performance.md
```diff
@@ -0,0 +1,142 @@
...
+Note: on a cluster you usually have to load a CUDA/ROCm toolkit and point `CUDA_HOME` at it, otherwise Numba
+cannot compile the kernels even though the GPU itself is visible. See
+[CUDA runtime not found](#numba-does-not-find-the-cuda-runtime-cudais_available-is-false) below.
+
+It is recommended to verify your cuda setup by running:
+
+```sh
+python -c "from numba import cuda; print(cuda.is_available())"
+# expected output: True
+```
```
BenjaminRodenberg: Maybe put this somewhere close to the `pip install -e .[gpu]`?

docs/Performance.md
```diff
@@ -0,0 +1,142 @@
...
+-->
+
+# Performance and Profiling Runs
+
+## Installation for Performance Runs
+
+Install EBFM with the performance features:
+
+```sh
+pip install -e .[performance]
```
BenjaminRodenberg:
```suggestion
pip install .[performance]
```

The `-e` is mainly for development. This guide is more intended for production cases; if developers want to use it I assume they will add the `-e` on their own.

Instead of `pip install .` you could also use `pip install ebfm`; Again: I would expect users of this guide to get EBFM from PyPI and developers should know what to do (we explain the different installation options elsewhere).

docs/Performance.md
```diff
@@ -0,0 +1,142 @@
...
+### GPU offloading: `--with-gpu`
+
+You can use this option when the node you run on has a GPU: it offloads the same kernels to the device. How much that
+gains depends on the mesh size and the GPU, and for small meshes it is not necessarily faster than `--with-numba`.
+`--with-gpu` and `--with-numba` are alternative backends for those kernels and therefore mutually exclusive.
+
+Install the GPU dependencies and run with the `--with-gpu` flag:
+
+```sh
+pip install -e .[gpu]
```
BenjaminRodenberg:
```suggestion
pip install ebfm[gpu]
```
see above

docs/Performance.md
```diff
@@ -0,0 +1,142 @@
...
+```sh
+python -c "from numba import cuda; print(cuda.is_available())"
+# expected output: True
+```
+
+At startup `--with-gpu` reports which stack and device it found, so you can check it went to the GPU you expected:
+
+```
+[GPU] backend enabled (nvidia). Device: NVIDIA A100-SXM4-80GB  free=79.15 GiB  total=79.15 GiB
+```
```
BenjaminRodenberg:
````suggestion
When `--with-gpu` is set EBFM will report at startup which stack and device has been found, so you can check it went to the GPU you expected:

```
[GPU] backend enabled (nvidia). Device: NVIDIA A100-SXM4-80GB  free=79.15 GiB  total=79.15 GiB
```
````

Is "stack" a common term? What's the stack in this message?

File: docs/Performance.md [hunk 22]
```diff
@@ -0,0 +1,142 @@
...
+
+## Timing Your Run
+
+To measure the total runtime, simply prepend your command with
```
BenjaminRodenberg:
```suggestion
To measure the total runtime, prepend your command with
```

docs/Performance.md
```diff
@@ -0,0 +1,142 @@
...
+ebfm --matlab-mesh examples/dem_and_mask.mat --dump-reference reference_run.npz
+```
+
+Then compare with:
+
+```sh
+python tools/compare_snapshots.py reference_run.npz new_run.npz
+```
+
+Note: If you use the random-forcing within EBFM in your testcases, make sure to additionally set the option `--random-seed`, as explained below.
```
BenjaminRodenberg:
```suggestion
Note: If you use the random-forcing within EBFM (`--which-option`) in your testcases, make sure to additionally set the option `--random-seed`, as explained below.
```
I would explicitly point to the option here. Don't know right now what this should be. So added the placeholder above.

BenjaminRodenberg (approved):
Looks good. Feel free to merge. I added a few more comments. But I don't think there are critical things left and we are really optimizing here. So feel free to discard anything that does not clearly convince you.

docs/Performance.md
matthias-ibach: I mean this is a fair point: other than the dependency, there is no real need to keep it off by default.   
The comparison runs show the NumPy and Numba paths agree within tolerance and the compile cost is smaller than I suggested. I double-checked in my benchmarks. It states that the Numba init was within ~0.05 s of the NumPy one (the kernels are compiled with cache=True, so it is a one-off per machine), while the time loop was 3–6 times faster. So it pays off almost immediately and the more so for the longer runs EBFM is actually used for.

I would still not address this in this MR, so I opened a follow-up issue to discuss it: #162.

docs/Performance.md
matthias-ibach: No, you are correct, it is a bit odd to start the sentence like this. I went ahead with your second suggestion.

docs/Performance.md
matthias-ibach: I agree with you. I would also expect users to get EBFM from PyPI, so I went ahead with `pip install ebfm[performance]`.

docs/Performance.md
matthias-ibach: Fair point, I repeated it here.

matthias-ibach: Done.

docs/Performance.md
matthias-ibach: Agreed and changed.

docs/Performance.md
matthias-ibach: Done.

docs/Performance.md
matthias-ibach: Changed.

docs/Performance.md
matthias-ibach: I added:
"Example: Run EBFM on the MATLAB example mesh with the Numba kernels on two threads:"

docs/Performance.md
matthias-ibach: Yes, that is a good point. Done, `pip install ebfm[gpu]` now sits next to the `[performance] install` and the GPU section only shows how to run it.

docs/Performance.md
matthias-ibach: Moved it up next to the install as well, together with the `CUDA_HOME` note that explains a False result.

docs/Performance.md
matthias-ibach: I dropped the "NVIDIA/AMD/--gpu-vendor" paragraph. The `--gpu-vendor` mechanics are indeed better suited for the `--help` section. But that AMD works at all, via numba.hip, isn't something a user would think to check `--help`. So I cut it to one sentence and keep the discovery bit:

"The same kernels run on NVIDIA and AMD GPUs; see --gpu-vendor in ebfm --help."

docs/Performance.md
matthias-ibach: In "[gpu] backend enabled" (NVIDIA), "stack" means which GPU software stack numba is using: `numba.cuda` for NVIDIA/CUDA or `numba.hip` for AMD/ROCm. Since the message prints the vendor, I changed it to:

"When --with-gpu is set, EBFM will report at startup which vendor backend (nvidia or amd) and which device have been found, so you can check it went to the GPU you expected:"

matthias-ibach (commented):
One remark: I added (and tested/verified this on the GPU) a warning suppresion of the probe-test and added a comment here.  
I know that this is not directly scope of this documentation only MR, but I simply did not want to add another MR round here (or a separate one either). Please feel free to review it if you have any objections to it.

